PR #36, opened 2023-08-19
feat: Installation of cross-env for 'yarn start' command on windows

Bobinstein:
installed cross-env as a dev dependency and updated the 'yarn start' script to eliminate syntax errors when run on a windows machine.


djwhitt:
Merged. Thanks!

PR #38, opened 2023-08-24
docs: updated open-api docs, .gitattributes

Bobinstein:
Expanded open-api docs and modified .gitattributes for new files.

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -48,7 +48,16 @@ tags:
   - name: Index Querying
     description: |
       Get data from the AR.IO Gateway index using GQL
+  - name: Admin
+    description: |
+      Access several password protected features and functions specific to your AR.IO Gateway.\nInput your ADMIN_API_KEY using the green "Authorize" button at the top of this page, or by clicking on the "lock" icon next to any password protected endpoint.
 components:
+  securitySchemes:
+    bearerAuth:
+      type: http
+      scheme: bearer
+      bearerFormat: userPassword
+      description: ADMIN_API_KEY set in your .env file.
   schemas:
     Block:
       type: object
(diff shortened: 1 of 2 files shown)


docs/openapi.yaml
```diff
@@ -48,7 +48,19 @@ tags:
   - name: Index Querying
     description: |
       Get data from the AR.IO Gateway index using GQL
+  - name: Admin
+    description: |
+      Access several password protected features and functions specific to your AR.IO Gateway.  
+
+
+      Input your ADMIN_API_KEY using the green "Authorize" button at the top of this page, or by clicking on the "lock" icon next to any password protected endpoint.
```
djwhitt, line 56: Good tip, but probably something that's more appropriate in an admin guide. The API docs should remain client agnostic.

docs/openapi.yaml
```diff
@@ -642,3 +691,85 @@ paths:
...
+        '401':
+          description: Unauthorized, API key likely incorrect.
+      security:
+        - bearerAuth: []
+  '/ar-io/admin/block-data':
+    put:
+      tags: [Admin]
+      summary: Blocks transactions or data-items so your AR.IO Gateway will not serve them.
+      description: | 
+        Submits a TX ID or data-item hash for content you do not want your AR.IO Gateway to serve. Once submitted, your Gateway will not respond to requests for these transactions or data-items.
```
djwhitt, line 742: `TX ID or data-item hash` should be `TX ID/data-item ID or sha-256 content hash`.

docs/openapi.yaml
```diff
@@ -48,7 +48,19 @@ tags:
...
+      Access several password protected features and functions specific to your AR.IO Gateway.  
+
+
+      Input your ADMIN_API_KEY using the green "Authorize" button at the top of this page, or by clicking on the "lock" icon next to any password protected endpoint.
 components:
+  securitySchemes:
+    bearerAuth:
+      type: http
+      scheme: bearer
+      bearerFormat: userPassword
```
djwhitt, line 62: I don't think `userPassword` is quite right here. It should probably be something like `apiToken`.

djwhitt (changes requested):

djwhitt (approved):

djwhitt:
Merged these manually. I'll try to fix the PR checks soon so I don't need to do that anymore.

PR #40, opened 2023-08-31
docs: update api docs for queue-tx

Bobinstein:


docs/openapi.yaml
```diff
@@ -706,8 +706,8 @@ paths:
   '/ar-io/admin/queue-tx':
     post:
       tags: [Admin]
-      summary: Prioritize a transaction in the queue.
-      description: Stage a specific TX ID as priority for your Gateway to locate and index.
+      summary: Prioritize a transaction or bundle in the queue.
+      description: Stage a specific TX ID as priority for your Gateway to locate and index. This can be any L1 transaction, or bundle if your Gateway is configured to unbundle and index.
```
djwhitt, line 710: I worry this will be confusing since you can't actually queue a nested bundle using this endpoint. Maybe we should keep the original wording, but add something like "this will trigger bundle processing if the transaction the TX is a bundle".

docs/openapi.yaml
```diff
@@ -719,7 +719,7 @@ paths:
                 id:
                   type: string
                   pattern: '^[0-9a-zA-Z_-]{43}$'
-                  description: TX ID of the transaction you would like to prioritize.
+                  description: TX ID of the transaction or bundle you would like to prioritize.
```
djwhitt, line 722: We should keep the original wording on this one. It must be a TX ID. An ID for a nested bundle wouldn't work.

djwhitt (changes requested):

djwhitt (approved):

djwhitt:
Looks like there are rebase conflicts. I'll sort those out and merge in a bit.

PR #41, opened 2023-08-31
feat: refactor 'nix develop' flake

hlolli:
using the legacy nixpkgs caused an issue on my aarch64-darwin system, as it tried to execute linux binaries, therefore I pass the system variable into the import of nixpkgs, and the issue is gone.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,22 +1,57 @@
 {
   "nodes": {
-    "nixpkgs": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
       "locked": {
-        "lastModified": 1648219316,
-        "narHash": "sha256-Ctij+dOi0ZZIfX5eMhgwugfvB+WZSrvVNAyAuANOsnQ=",
-        "owner": "NixOS",
-        "repo": "nixpkgs",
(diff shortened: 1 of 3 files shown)


djwhitt (changes requested):
Change itself looks fine, but I have two requests for the commit message:

- Switch it from `feat:` to `refactor:`
- Include the PR description in the body of the commit

I want to make sure we have a commit history that provide enough info for people to fairly easily make sense of our changes in the future.

djwhitt:
Force pushing to the branch is fine to update the commit message.

djwhitt:
I probably should expand slightly for clarity:

I overall don't care that much about conventional commit style, but I think it makes sense to be somewhat strict about `feat` and `fix` usage since those are useful for constructing change logs and release notes. That's a relatively minor consideration though, the thing that is much more important to me is explaining why changes are made somewhere in the commit message. It helps so much when trying to make sense of the history later. In OSS projects that's especially important. Of course, there are exceptions (e.g. a series of related refactoring might not need to re-explain the rationale for each change), but those should be rare.

hlolli:
wait let me remove those leading whitespaces in my commit message....

djwhitt:
@hlolli Looks good. Appreciate it! :pray:

Rebased and merged into `develop`.

PR #42, opened 2023-08-31
feat(apollo): disable persistedQueries

hlolli:
This should disable the warning (by addressing the issue)

```
...
Your server is vulnerable to denial of service attacks via memory exhaustion.
...
```


djwhitt (changes requested):
Change itself looks good, but I have two requests for the commit message:

- Switch it from `feat:` to `fix:`
- Include a description of the reason for the change (maybe also a link to Apollo docs) in the commit message.

Force pushing to the branch is fine to update the commit message.

PR #43, opened 2023-09-01
fix(ci): lcov-reporter-action

hlolli:
* prevent hitting body payload limit in test-reporter
* set 'filter-changed-files' to true to only check changed files


djwhitt:
Appreciate work on this one, but closing since we can use Codecov now.

Issue #26, opened 2023-08-09
feat: path manifest fallback to `*`


twilson63:
BACKGROUND: https://specs.g8way.io/#/view/lXLd0OPwo-dJLB_Amz5jgIeDhiOkjXuM3-r0H_aiNj0

Modern js frameworks rely on push state, this is a feature where if the path request is not found, it will return the `index` file. This allows modern js frameworks to handle routing and hyperlinks in a web-native way, with no `#` hash routing required.

If gateways could handle a `not_found` request for any route from a path manifest TX id with the result returning the '*' property of the manifest, we could support many modern js frameworks out of the box with zero workarounds.

This issue is a request to build such a feature in middleware so that it does not impact the development of the gateway and can be installed by any gateway operator.

Here is an example:

```json
{
  "manifest": "arweave/paths",
  "version": "0.1.0",
  "index": {
    "path": "index.html"
  },
  "paths": {
    "index.html": {
      "id": "cG7Hdi_iTQPoEYgQJFqJ8NMpN4KoZ-vH_j7pG4iP7NI"
    },
    "js/style.css": {
      "id": "fZ4d7bkCAUiXSfo3zFsPiQvpLVKVtXUKB6kiLNt2XVQ"
    },
    "css/style.css": {
      "id": "fZ4d7bkCAUiXSfo3zFsPiQvpLVKVtXUKB6kiLNt2XVQ"
    },
    "css/mobile.css": {
      "id": "fZ4d7bkCAUiXSfo3zFsPiQvpLVKVtXUKB6kiLNt2XVQ"
    },
    "assets/img/logo.png": {
      "id": "QYWh-QsozsYu2wor0ZygI5Zoa_fRYFc8_X1RkYmw_fU"
    },
    "assets/img/icon.png": {
      "id": "0543SMRGYuGKTaqLzmpOyK4AxAB96Fra2guHzYxjRGo"
    },
    "*": {
      "id": "cG7Hdi_iTQPoEYgQJFqJ8NMpN4KoZ-vH_j7pG4iP7NI"
   }
  }
}
```

djwhitt:
[...]

PR #46, opened 2023-09-08
feat: Support for wildcards

angrymouse:


src/lib/encoding.ts
```diff
@@ -324,29 +340,40 @@ export function resolveManifestStreamPath(
 ): Promise<string | undefined> {
   return new Promise((resolve, reject) => {
     const emitter = parseManifestStream(stream);
-
+    let resolved=false
```
arielmelendez, line 343: Written to but never read?

angrymouse:
As per https://github.com/ar-io/ar-io-node/issues/26, however without "overall" improvement of just allowing to put id into index, only wildcard resolving part

src/lib/encoding.ts
```diff
@@ -295,6 +296,18 @@ export function parseManifestStream(stream: Readable): EventEmitter {
         paths = {};
       }
     }
+    if (
```
arielmelendez, line 299: Please try to keep the code style of the surrounding code blocks, i.e. a blank line followed by a comment that highlights what the subsequent conditional block is intending to select on during parsing, followed by the conditional block itself.

src/lib/encoding.ts
```diff
@@ -231,6 +231,7 @@ export function parseManifestStream(stream: Readable): EventEmitter {
   let currentKey: string | undefined;
   const keyPath: Array<string | number> = [];
   let indexPath: string | undefined;
+  let wildcardpath: string | undefined;
```
arielmelendez, line 234: Nit: keep variable casing consistent with rest of the codebase, i.e. `wildcardPath`

arielmelendez: Actually maybe this is an error since you're using `wildcardPath` below but `wildcardpath` here? 🧐

arielmelendez:
Thanks for the PR! I'm kicking off some discussion internally around whether to accept this PR pending a few decisions, e.g.
- should we accept a PR that only partially implements the spec updates outlined here: https://github.com/ar-io/ar-io-node/issues/26#issuecomment-1688455200
- should a partial or full implementation of said spec updates require stricter manifest version checking
- are there tests that need to be added/created

angrymouse:
Let me put my 2 pennies.
Regarding 1st, I don't think it has to be 1 PR, due to how different points in this summary are - for example, 256kb limit has *nothing* to do with wildcard support except that it may optimize it.
Also I don't think it requires new manifest version either, as it doesn't overwrite what's already done and isn't breaking change. Replacing "path" to "id" would be breaking change, but not adding wildcard.
Tests are up to you

djwhitt:
@angrymouse Really appreciate you digging in on this! My initial thoughts:

- A version bump is important because it communicates to those parsing the manifest what they can expect to find in it. It's difficult to scale manifest processing if gateways have to parse the entire manifest to make that determination. Gateways can also potentially use the version number to communicate to their users what features they support.
- The 256kb limit isn't something that needs to be implemented as part of this. That's an expectation I think should be included in the associate spec updates though since it's hard to make promises about how quickly or whether manifests of unbounded size will be processed.
- We do not want to support paths for wildcards. Supporting paths is a major performance issue for on-demand parsing and ideally shouldn't have been part of the manifest spec in the first place. We could use the wildcard change as an opportunity to drop support for it for indexes too and make manifest processing simpler for gateways in the future. I think that's still up for discussion though. The main thing is that I do not think we should make the problem worse by supporting paths for wildcards.

tl;dr - great work so far, but let's switch to IDs for wildcards and include a version bump + check as part of this so that users can communicate their expectations about manifest features. We can continue discussion on whether we want to drop paths for index.

We'll also wants tests too, but we can dig into that after we're through discussing the functionality.

Oh, PRs should be against `develop` as described in the contributing doc (https://github.com/ar-io/ar-io-node/blob/main/CONTRIBUTING.md#pull-request-checklist), but that's just for future reference. This area isn't under heavy development, so we can easily rebase and merge into `develop` when it's ready.

angrymouse:
About 3rd, I don't think there will be a lot of issues specifically for wildcards in supporting paths (because wildcard is processed as the last thing, after every path was checked) @djwhitt

angrymouse:
Bumped version to 0.1.1, made resolving with only id possible

djwhitt:
> About 3rd, I don't think there will be a lot of issues specifically for wildcards in supporting paths (because wildcard is processed as the last thing, after every path was checked) @djwhitt

This is only true as long as we don't enforce path sorting. If in a future version we do as the original comment suggested and only support on-demand parsing with sorted paths, we can stop parsing early.

Also just to set expectations - there will probably be a little pause on this over the weekend, but we'll definitely pick back up next week and keep it moving.

djwhitt:
@angrymouse Apologies for the slow follow up this week. We were fighting some ops battles yesterday. I think we can move forward with this after a couple adjustments:

> Bumped version to 0.1.1

Let's bump to 0.2 since we're adding new functionality as opposed to fixing issues (we might as well follow semver unless we have a specific reason to do something different).

Also, now that we have consensus on the changes, go ahead and add tests to cover the wildcard functionality.

Finally, just so everyone following along is aware - once the dev work here is completed, there will probably still be a bit of a delay before this is available on arweave.net. For clarity, we may also want to update the ANS spec before rolling it out there.

angrymouse:
Hey @djwhitt ! Sorry didn't check out... Will do rn

angrymouse:
@djwhitt done

angrymouse:
Hey @djwhitt , any updates?

djwhitt:
@angrymouse Thanks for the tweaks! Just one more request on the tests - can you include a test for the case where the wildcard uses a `path` instead of an `id` since that's still supported in the code. We'll be good to merge after that I think.

djwhitt:
@angrymouse I just pulled this to see if I could add the requested test myself and discovered the test you added is failing. I'm happy to merge when that's debugged and the requested test is added.

angrymouse:
Thanks, I'm not good in writing tests

djwhitt:
To be clear, I was requesting that you debug and fix the test before we move on. I'm not actually sure certain the code is working in it's current state. Did you manually test it somehow?

angrymouse:
Hey @djwhitt , no, I didn't test it and I don't recall you asking for that. Also, to make it clear, it's just an open, enthusiastic, non-inventivized effort to help *your* software, meaning I don't really have to do the corporate logic like tests

angrymouse:
But the error seems to come from types, not the logic

djwhitt:
I'm not trying to make accusations. It was just unclear to me if the code works, so I was curious how or if you verified that yourself.

PRs are work for the maintainers. If we get a PR and we're uncertain that it's working it's going to take more effort to integrate it. We will still eventually integrate it assuming it's in line with the the project goals (this one is). It's just that the more work there is for us to do on the PR, the more it becomes subject to our (ar.io) priorities rather than the person making the PR.

> But the error seems to come from types, not the logic

That's just because we run build before tests in the actions. If you run the tests locally you'll see a different error. The first error you'll see is from malformed JSON in the test. If you fix that you'll see another error which I haven't dug into yet.

angrymouse:
Will look into fixing it, thanks for understanding

djwhitt:
@angrymouse Thanks! I know PR processes can be frustrating  too. I appreciate your patience as we continue to refine ours. :pray:

djwhitt:
@jim-toth Happy to have you dig in on testing this if you have time (since you called it out as blocking). Though, be sure to coordinate with @angrymouse to avoid duplicate effort.

djwhitt:
Closing. Release 14 includes support for `fallback` paths (see https://docs.ar.io/manifests/ for details).

Issue #47, opened 2023-09-10
docs: specify and describe all envs in docs


rixcian:
There's no documentation about the `envs` the gateway operator can use/edit.

The operator has to look at the `config.ts` file and find out what `envs` the gateway is trying to load from the `.env` file.

I would suggest creating the following:
- [ ] `.env.example` file, where all possible `envs` will be specified and describe what they do
- [ ] Specify and describe all `envs` in the clear table for these files:
  - [ ] README.md
  - [ ] Official docs (`linux-setup.md`, `windows-setup.md`)

The table in markdown could look like this:

| ENV_NAME     | DEFAULT_VALUE | DESCRIPTION                                    |
|--------------|---------------|------------------------------------------------|
| START_HEIGHT | 0             | Starting block height for node synchronization |
| ...          | ...           | ...                                            |

If you're okay with this proposal, could you assign it to me? I'm open to doing it.

Thanks in advance for feedback :)

rixcian:
I think this PR can be closed now (#49 is already merged to the main branch).

djwhitt:
@rixcian Closing. Thanks for the reminder!

PR #49, opened 2023-09-15
docs: envs documentation

rixcian:
This PR should resolve the #47 issue.

I've noticed that the ENVs are quite well documented in the official docs. So I've just created a `.env.example` file with all possible ENVs and their description.

I've also created an `env.md` file where all the ENVs are specified in the markdown table. It's up to you if you want to use the table in some different place in the docs.

I've heard that @Bobinstein is working on docs, so his feedback will be more than appreciated.


docs/envs.md
Bobinstein, line 1: you can add 
LOG_FORMAT  |  string   | "simple"  | sets the format of output logs, accepts "simple" and "json" 
INSTANCE_ID   |  string   | ""             | Adds an "instanceId" field to output logs

.env.example
Bobinstein, line 1: wrapping the values for ANS104_UNBUNDLE_FILTER and ANS104_INDEX_FILTER in single quotes can lead to parsing errors. The values do not need to be wrapped in any quotes.

docs/envs.md
rixcian: I've added the missing ENVs.

.env.example
rixcian: I've removed the single quotes.

djwhitt:
Thanks, @rixcian! Rebased and merged. Worth noting I made a couple of minor corrections in the process:

- https://github.com/ar-io/ar-io-node/commit/38210b9979d50ceb00a24feb12bd244b09f1d3c4
- https://github.com/ar-io/ar-io-node/commit/eb843a76cf7389fefac5df908021edc1a1a754b7

PR #51, opened 2023-09-18
fix(ci): this should be `/ar-io/healthcheck` right

Megumiiiiii:
```
ghcr.io/ar-io/ar-io-core:latest    "/bin/sh docker-entr…"   44 seconds ago   Up 40 seconds (healthy)   0.0.0.0:4000->4000/tcp                                      ar-io-node-core-1
```


djwhitt:
@Megumiiiiii Thanks for the PR! It's merged now. For future reference, please open PRs against `develop` instead of `main`.

PR #52, opened 2023-09-20
Add subdomain support

kay-is:
The sandbox and ArNS middleware didn't work when the node was hosted on a subdomain.

This PR adds subdomain support to both middlewares, so the node doesn't require a dedicated apex domain for hosting.

### DIFF
--- a/src/middleware/arns.ts
+++ b/src/middleware/arns.ts
@@ -18,6 +18,7 @@
 import { Handler } from 'express';
 import { asyncMiddleware } from 'middleware-async';
 
+import * as config from '../config.js';
 import { sendNotFound } from '../routes/data.js';
 import { NameResolver } from '../types.js';
 
@@ -25,24 +26,21 @@ const EXCLUDED_SUBDOMAINS = new Set('www');
 
 export const createArnsMiddleware = ({
   dataHandler,
-  rootHost,
   nameResolver,
 }: {
   dataHandler: Handler;
-  rootHost: string;
   nameResolver: NameResolver;
 }): Handler =>
   asyncMiddleware(async (req, res, next) => {
-    const rootHostSubdomainLength = rootHost.split('.').length - 2;
     if (
       // Ignore subdomains that are part of the ArNS root hostname.
       !Array.isArray(req.subdomains) ||
-      req.subdomains.length === rootHostSubdomainLength
+      req.subdomains.length === config.ROOT_HOST_SUBDOMAIN_LENGTH
     ) {
       next();
       return;
     }
-    const arnsSubdomain = req.subdomains[req.subdomains.length - 1];
+    const arnsSubdomain = req.subdomains[config.ROOT_HOST_SUBDOMAIN_LENGTH - 1];
     if (
       EXCLUDED_SUBDOMAINS.has(arnsSubdomain) ||
       // Avoid collisions with sandbox URLs by ensuring the subdomain length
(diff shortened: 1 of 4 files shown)


src/middleware/arns.ts
```diff
@@ -25,33 +25,42 @@ const EXCLUDED_SUBDOMAINS = new Set('www');
...
   dataHandler,
+  rootHost,
   nameResolver,
 }: {
   dataHandler: Handler;
+  rootHost: string;
   nameResolver: NameResolver;
 }): Handler =>
   asyncMiddleware(async (req, res, next) => {
+    const rootHostSubdomainLength = rootHost.split('.').length - 2;
```
djwhitt, line 36: @kay-is, let's reduce duplication by calculating this value in `config.ts` (create a  `ROOT_HOST_SUBDOMAIN_LENGTH` const) and then passing that into the middlewares. Alternately, you could create a utility function and use it in both locations, but I think the config constant approach is probably simpler for now.

src/middleware/sandbox.ts
```diff
@@ -21,9 +21,10 @@ import { base32 } from 'rfc4648';
 
 import { fromB64Url } from '../lib/encoding.js';
 
-function getRequestSandbox(req: Request): string | undefined {
-  if (req.subdomains.length === 1) {
-    return req.subdomains[0];
+function getRequestSandbox(req: Request, rootHost: string): string | undefined {
+  const rootHostSubdomainLength = rootHost.split('.').length - 2;
```
djwhitt, line 25: See duplication comment here: https://github.com/ar-io/ar-io-node/pull/52/files#diff-d44b648d161b4fbef365ace7f13470b7c322b08a5f5226533af45c2d14c709b0R36

src/middleware/arns.ts
```diff
@@ -25,33 +25,42 @@ const EXCLUDED_SUBDOMAINS = new Set('www');
...
+    }
+    const arnsSubdomain = req.subdomains[req.subdomains.length - 1];
+    if (
+      EXCLUDED_SUBDOMAINS.has(arnsSubdomain) ||
       // Avoid collisions with sandbox URLs by ensuring the subdomain length
       // is below the mininimum length of a sandbox subdomain. Undernames are
       // are an exception because they can be longer and '_' cannot appear in
       // base32.
-      (req.subdomains[0].length <= 48 || req.subdomains[0].match(/_/))
+      (arnsSubdomain.length > 48 && !arnsSubdomain.match(/_/))
```
djwhitt, line 52: Nice bit of conditional cleanup here. :slightly_smiling_face:

djwhitt (changes requested):
Thanks for the PR! Just one request re reducing duplication, otherwise this looks great.

src/middleware/arns.ts
kay-is: I thought about doing this, but I didn't want to contaminate app.js with implementation details of the middleware. If you okay with that, I'll change it.

kay-is:
I switched it to the config and also cleaned up the middleware parameters.

djwhitt (approved):

djwhitt:
@kay-is One final note (didn't notice this earlier) - PRs should be against `develop` as described in [CONTRIBUTING.md](https://github.com/ar-io/ar-io-node/blob/main/CONTRIBUTING.md). NBD though, appreciate the PR regardless. :slightly_smiling_face:

PR #63, opened 2023-11-20
docs(Diagrams): Added puml diagrams.  PE-4847

Bobinstein:


docs/diagrams/src/Gateway.puml
```diff
@@ -13,35 +13,27 @@ LAYOUT_TOP_DOWN()
...
     Component(observer, "Observer Service")
 
-    System(storage, "Local Storage"){
-        ContainerDb(chunks, "Chunk data")
-        ContainerDb(headers, "Header Data")
-        ContainerDb(contiguous, "Contiguous Data")
-        ContainerDb(indexes, "Index Data")
-        ContainerDb(reports, "Observer Reports")
-    }
+    ContainerDb(index, "Chain and Bundle Index", "Chunked, header, and contiguous data and Observer reports")
```
djwhitt, line 23: Oh, sorry, I wasn't specific enough in my comment. I was only trying to suggest that "Index Data" should be renamed to "Chain and Bundle Index" not that the entire group should be combined into one item.

djwhitt (commented):
@Bobinstein You removed a little more detail in one place than I intended (see inline comment). I should have been more specific in my wording there. All the other changes look good. Add back some of that original detail along with the suggested wording change and I think this is good to go.

PR #66, opened 2023-12-05
fix(ans-104): ensure that ans-104 streams are destroyed on error PE-5088

djwhitt:
With ans-104 unbundling enabled, we've occasionally observed a large increase in disk usage that disappears after the gateway service is stopped. Our best guess is that this is related to hanging on to bundle temporary file handles. This change attempts to address that by ensuring temporary file streams are always destroyed when errors occur.


djwhitt:
Hard to say for sure this is resolved, but I'm merging now since testing on the branch looks promising (just stopped the service on a system with 12 unbundle threads and noticed zero change in disk usage after the service was stopped).

PR #71, opened 2023-12-17
rename created_at to indexed_at in DB

mskutlu:


djwhitt:
Thanks for the PR! Given everything else happening right now, I probably won't get back to this one till the new year, but we will integrate it soon. It actually helps out with something I'm currently working on too. :slightly_smiling_face:

djwhitt:
I ended up needing to take a little bit different approach than what's in this PR (see https://github.com/ar-io/ar-io-node/commit/a0eece04399eaee04c442b588650a970d33ec609 for details if you're curious), but I appreciate your desire to contribute and the time you put into this!

PR #72, opened 2023-12-22
feat(webhooks): Added webhooks based on ANS104_INDEX_FILTER

Bobinstein:
Added webhooks to operator specified server, set as `WEBHOOK_TARGET_SERVER` in .env

Example server to receive webhooks can be found [here](https://github.com/Bobinstein/listener)


t8:
Will be great to see this merged. Eager to ship notifications in ArConnect!

cc @7i7o @nicholaswma

samcamwilliams:
Haven't tried this PR in practice, but in principle this sounds awesome! Would love to see it in the gateways, as a ton of services on the permaweb could be made more efficient with subscriptions rather than polling.

Thank you for contributing, @Bobinstein 🫡.

How does it look, ARIO crew?
Sam

twilson63:
Looking forward to it as well! Was this part of hack week?

PR #82, opened 2024-01-26
chore(PE-5507): bump observer to latest image and add warp cache vol…

dtfiedler:
…ume to docker-compose

This version of the observer uses lmdb as its contract store for improved performance. The volume allows the observer to persist cache accross restarts

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -85,7 +85,7 @@ services:
     volumes:
       - ${TEMP_DATA_PATH:-./data/tmp}:/app/data/tmp
       - ${REPORTS_DATA_PATH:-./data/reports}:/app/data/reports
-      - ${WARP_CACHE_PATH:./cache/warp}:/app/cache/warp
+      - ${WARP_CACHE_PATH:-./cache/warp}:/app/cache/warp
       - ${WALLETS_PATH:-./wallets}:/app/wallets
     environment:
       - PORT=5000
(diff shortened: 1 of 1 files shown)


docker-compose.yaml
```diff
@@ -78,13 +78,14 @@ services:
...
   observer:
-    image: ghcr.io/ar-io/ar-io-observer:${OBSERVER_IMAGE_TAG:-39783cb6ab7586a443b0b3907e18b5cca914e72d}
+    image: ghcr.io/ar-io/ar-io-observer:${OBSERVER_IMAGE_TAG:-5bad8c827fe19158f61c4aab090e37899b5e5328}
     restart: on-failure:5
     ports:
       - 5000:5000
     volumes:
       - ${TEMP_DATA_PATH:-./data/tmp}:/app/data/tmp
       - ${REPORTS_DATA_PATH:-./data/reports}:/app/data/reports
+      - ${WARP_CACHE_PATH:-./cache/warp}:/app/cache/warp
```
djwhitt, line 88: Let's put this under `./data/warp/cache` since that's where other mounted volumes live.

djwhitt (approved):

djwhitt (changes requested):
Clicked the wrong button initially. :sweat: See my comment about the data path.

docker-compose.yaml
dtfiedler: changed the volume path in [e245e9c](https://github.com/ar-io/ar-io-node/pull/82/commits/e245e9cea99d638c3569ecb7e7c5995a5fb52b0b)

as for the app path - i'll have to make a commit to the observer to not use the default cache path

djwhitt: Eh, I'm not worried about inside the container.

djwhitt (approved):

PR #84, opened 2024-02-01
PE-5192:  data download worker

djwhitt:
Optimizes ANS-104 processing by separating the initial ANS-104 bundle download from its subsequent unbundling. This enables independent scaling of download (network IO) and unbundling (disk IO and compute) to better utilize available resources. The number of parallel downloads can be set using the ANS104_DOWNLOAD_WORKERS environment variable.

For the time being, unbundling is queued directly without after the data download is complete. In the future we may decouple data downloads from unbundling using events and add support for a download filter that differs from the unbundling filter.


src/workers/bundle-data-importer.ts
```diff
@@ -0,0 +1,128 @@
...
+      Math.max(workerCount, 1), // fastq doesn't allow 0 workers
+    );
+  }
+
+  async queueItem(
+    item: UnbundleableItem,
+    prioritized: boolean | undefined,
+  ): Promise<void> {
+    if (this.workerCount === 0) {
+      return;
```
karlprieb, line 83: I think it would be good to have a debug log for this case.

karlprieb (approved):

PR #85, opened 2024-02-01
feat(filters): add NegateMatch and not filter

karlprieb:
Add NegateMatch. A match that negates all matches.
Filter can use `not` to negate a match.


djwhitt (approved):

vilenarios:
@coderabbitai full review

Please re-review — the previous review was on the initial commit only. Since then we've addressed your two Major findings in eb14260 (default NETWORK_SOURCE='ao' + parsePositiveIntEnv/parseNonNegativeFloatEnv helpers for the cranker numeric env vars).

PR #86, opened 2024-02-07
feat(system): graceful shutdown

karlprieb:
Implement `stop` methods to clear queues and stop workers in the following components:
  - standalone-sqlite
  - ans-104 lib
  - ans104-data-indexer
  - ans104-unbundler
  - bundle-data-importer
  - bundle-repair-worker
  - data-item-indexer
  - transaction-fetcher
  - transaction-importer
  - transaction-offset-importer
  - transaction-offset-repair-worker
  - transaction-repair-worker
  - webhook-emitter

Through a `shutdown` function it will close express server and wait for all these componenets/workers to stop.


djwhitt (approved):
Looks good! This should make shutdown a lot cleaner.

PR #92, opened 2024-03-04
feat(data handlers): stop processing data stream for HEAD requests

karlprieb:
When doing a HEAD request for `/tx-id` or `/raw/tx-id` endpoints the server will not process the data stream while responding. It will use the cached data information to set the headers and if no cache is found it will process the data from the referenced gateway and cache it.

### DIFF
--- a/src/routes/data/handlers.ts
+++ b/src/routes/data/handlers.ts
@@ -127,6 +127,8 @@ const handleRangeRequest = (
 
     if (req.method === REQUEST_METHOD_HEAD) {
       res.end();
+      data.stream.destroy();
+      return;
     }
 
     // Create a custom Transform stream to filter the range
@@ -264,6 +266,8 @@ export const createRawDataHandler = ({
 
         if (req.method === REQUEST_METHOD_HEAD) {
           res.end();
+          data.stream.destroy();
+          return;
         }
 
         data.stream.pipe(res);
@@ -360,6 +364,7 @@ const sendManifestResponse = async ({
 
         if (req.method === REQUEST_METHOD_HEAD) {
           res.end();
+          data.stream.destroy();
           return true;
         }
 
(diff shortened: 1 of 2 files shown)


src/routes/data/handlers.ts
```diff
@@ -248,12 +255,17 @@ export const createRawDataHandler = ({
...
-        handleRangeRequest(log, rangeHeader, res, data, dataAttributes);
+        handleRangeRequest(log, rangeHeader, res, req, data, dataAttributes);
       } else {
         // Set headers and stream data
         setDataHeaders({ res, dataAttributes, data });
         setRawDataHeaders(res);
         res.header('Content-Length', data.size.toString());
+
+        if (req.method === REQUEST_METHOD_HEAD) {
+          res.end();
```
djwhitt, line 266: We should destroy the stream and return early here.

src/routes/data/handlers.ts
```diff
@@ -122,6 +125,10 @@ const handleRangeRequest = (
         DEFAULT_CONTENT_TYPE,
     );
 
+    if (req.method === REQUEST_METHOD_HEAD) {
+      res.end();
```
djwhitt, line 129: We should destroy the stream and return early here.

src/routes/data/handlers.ts
```diff
@@ -515,6 +533,11 @@ export const createDataHandler = ({
           data,
         });
         res.header('Content-Length', data.size.toString());
+
+        if (req.method === REQUEST_METHOD_HEAD) {
+          res.end();
```
djwhitt, line 538: We should destroy the stream and return early here.

src/routes/data/handlers.ts
```diff
@@ -345,6 +357,12 @@ const sendManifestResponse = async ({
...
```
djwhitt, line 362: We should destroy the stream here.

djwhitt (changes requested):

djwhitt (approved):

PR #99, opened 2024-03-22
feat(PE-5859): support standalone arns-resolver

dtfiedler:
Implements the `NameResolver` interface in the `StandaloneArNSResolver` class whucg  arns name data from locally hosted arns-resolver service. Continues to use gateway by default - but allows for configurability via two new environment variables to select which resolver service to use. The changes do not remove any env vars or change default behavior. Additionally, the APIs on the arns-resolver service follow existing patterns and can be added to the envoy template once testing validates the expected behavior.

### DIFF
--- a/src/init/resolvers.ts
+++ b/src/init/resolvers.ts
@@ -1,0 +1,50 @@
+/**
+ * AR.IO Gateway
+ * Copyright (C) 2022-2023 Permanent Data Solutions, Inc. All Rights Reserved.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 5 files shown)


src/init/resolvers.ts
```diff
@@ -0,0 +1,50 @@
...
+import { TrustedGatewayArNSResolver } from '../resolution/trusted-gateway-arns-resolver.js';
+import { NameResolver } from '../types.js';
+
+export const createArNSResolver = ({
+  log,
+  type,
+  url,
+}: {
+  log: Logger;
+  type: string;
```
karlprieb, line 29: can we have a better type like `'resolver' | 'gateway'`?

dtfiedler: bc it's provided as an environment variable - we'd have to force the type using `as` when calling the function [here](https://github.com/ar-io/ar-io-node/pull/99/files#diff-6fe6a66d9e19964283ad8fcf5ad1a9bf0e8a32a22124ef6f28474d92fda574edR395-R396) - which makes the type somewhat useless - the `switch` statement handles invalid values appropriately but at runtime.

TLDR - we could set this to `'resolver' | 'gateway' | string` to avoid forcing types on env vars

djwhitt (approved):
:+1: Nice and simple.

PR #101, opened 2024-03-28
feat(data-handlers): add X-Cache header

karlprieb:


src/data/read-through-data-cache.ts
```diff
@@ -132,13 +133,15 @@ export class ReadThroughDataCache implements ContiguousDataSource {
...
     );
+
     if (cacheData !== undefined) {
       return {
         hash: attributes?.hash,
         stream: cacheData.stream,
         size: cacheData.size,
-        sourceContentType: dataAttributes?.contentType,
-        verified: dataAttributes?.verified ?? false,
+        sourceContentType: attributes?.contentType,
```
karlprieb, line 142: @djwhitt I believe the intent was to use `attributes` instead of `dataAttributes` here, right?

karlprieb:
@djwhitt should we set `const size = parentData.size ?? dataSize;` as we're returning the parent data here?
https://github.com/ar-io/ar-io-node/blob/169d2c7551dc9f86a8a5b7e4406e06482c73f5c6/src/data/read-through-data-cache.ts#L105-L106

src/data/read-through-data-cache.ts
djwhitt: Yup. Good catch. This looks like a mistake.

djwhitt:
> @djwhitt should we set `const size = parentData.size ?? dataSize;` as we're returning the parent data here?
> 
> https://github.com/ar-io/ar-io-node/blob/169d2c7551dc9f86a8a5b7e4406e06482c73f5c6/src/data/read-through-data-cache.ts#L105-L106

This one is intentional. I think in theory either will work, but the goal is to treat the `dataAttributes` as authoritative.

PR #103, opened 2024-04-04
docs: add comments describing events PE-5900

djwhitt:
Groups events and adds descriptive comments for them. JSDoc style comments are used so that they show up as LSP hover text. @constant annotations were omitted since they seem redundant, but we can add them though if there's utity in doing so.


dtfiedler (approved):
📝

karlprieb (approved):

PR #104, opened 2024-04-09
feat(nodejs): upgrade to nodejs v20

karlprieb:
- upgrades the project to nodejs v20
- migrate test runner from mocha/chai to nodejs native test runner


Dockerfile
```diff
@@ -14,7 +14,7 @@ RUN yarn install \
     && yarn install --production
 
 # Runtime
-FROM gcr.io/distroless/nodejs${NODE_VERSION_SHORT}-debian11
+FROM gcr.io/distroless/nodejs${NODE_VERSION_SHORT}-debian12
```
karlprieb, line 17: @djwhitt are you ok using debian12?

package.json
```diff
@@ -96,6 +95,10 @@
     "ts-sinon": "^2.0.2",
     "typescript": "^4.7.3"
   },
+  "resolutions": {
+    "gc-stats/nan": "2.18.0",
+    "express-prometheus-middleware/prometheus-gc-stats/gc-stats": "1.4.1"
+  },
```
karlprieb, line 101: It was required to compile it in the newer v8 from nodejs 20.

We need to move on from prometheus-gc-stats as it's not maintained since 3 years ago.

src/data/gateway-data-source.test.ts
```diff
@@ -15,82 +15,83 @@
...
-import sinonChai from 'sinon-chai';
 import * as winston from 'winston';
 import { GatewayDataSource } from './gateway-data-source.js';
 
-chai.use(sinonChai);
+let log: winston.Logger;
+let dataSource: GatewayDataSource;
+let mockedAxiosInstance: any;
 
-describe('GatewayDataSource', function () {
```
dtfiedler, line 27: native testrunner doesn't support describe blocks?

package.json
dtfiedler: this isn't a package - but should we but the required node version in the package.json?

```json
  "engines": {
    "node": ">=20"
  },
```

src/data/gateway-data-source.test.ts
karlprieb: It supports, it's down below on line 59

PR #108, opened 2024-04-24
feat: MempoolWatcher index pending txs

karlprieb:


src/app.ts
```diff
@@ -41,6 +41,7 @@ if (config.START_WRITERS) {
   system.txOffsetRepairWorker.start();
   system.txRepairWorker.start();
   system.bundleRepairWorker.start();
+  system.mempoolWatcher?.start();
```
karlprieb, line 44: I think it make sense to just start it if `config.START_WRITERS` is `true`, right?

src/config.ts
```diff
@@ -209,6 +209,18 @@ export const TRUSTED_ARNS_RESOLVER_URL = env.varOrDefault(
...
+//
+// Mempool watcher
+//
+//
+export const ENABLE_MEMPOOL_WATCHER =
+  env.varOrDefault('ENABLE_MEMPOOL_WATCHER', 'false') === 'true';
+
+export const MEMPOOL_POLLING_INTERVAL_MS = +env.varOrDefault(
+  'MEMPOOL_POLLING_INTERVAL_MS',
+  '300000', // 5 minutes
```
karlprieb, line 221: is 5 minutes a good default value?

src/workers/mempool-watcher.ts
```diff
@@ -0,0 +1,102 @@
...
+  }) {
+    this.log = log.child({ class: 'MempoolWatcher' });
+    this.arweaveClient = arweaveClient;
+    this.txFetcher = txFetcher;
+    this.mempoolPoolingIntervalMs = mempoolPoolingIntervalMs;
+    this.shouldRun = false;
+  }
+
+  private normalizePendingTxs(txs: string[]): void {
+    const mempoolTxs = new Set(txs);
```
karlprieb, line 55: I transformed it to a `Set` because it's faster to iterate as we don't know the size of this list, but maybe this is not required

djwhitt: I'm fine either way on this one. I suspect an array actually ends up being faster, but I don't think we should worry about it too much.

djwhitt (commented):
I'll wait for tests before clicking approve, but this looks good. The code is pleasingly simple.

src/workers/mempool-watcher.ts
karlprieb: My concern with this `Set` is memory as we keep the array and the Set for a small period of time. So maybe just iterating over the array would be better considering memory usage

PR #109, opened 2024-04-24
Fix CodeDeploy BeforeInstall script

kay-is:
- Calling docker-compose before deleting `docker-compose.yaml`
- Checking that DB files exist before attempting a copy


djwhitt (approved):
:+1: Thanks! Never expected to get PRs on these.

PR #110, opened 2024-04-25
WIP: litestream sqlite backup service

hlolli:
Some experimentations on the docker-compose config and configuration setup for easy configuration integration.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,5 +1,10 @@
----
-version: '3.0'
+version: '3.8'
+
+volumes:
+  chunks-data:
+  contiguous-data:
+  headers-data:
+  db-data:
 
 services:
   envoy:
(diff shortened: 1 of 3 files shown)


docker-compose.yaml
```diff
@@ -1,5 +1,10 @@
----
-version: '3.0'
+version: '3.8'
+
+volumes:
+  chunks-data:
+  contiguous-data:
+  headers-data:
+  db-data:
```
djwhitt, line 7: I like this structure, but we need to continue to use volume mounts specified by environment variables by default for these.

litestream.yaml
```diff
@@ -0,0 +1,13 @@
+dbs:
```
djwhitt, line 1: Checkout how our Envoy container renders a template on startup: https://github.com/ar-io/ar-io-node/blob/develop/envoy/docker-entrypoint.sh#L12

docker-compose.yaml
```diff
@@ -1,5 +1,10 @@
----
-version: '3.0'
+version: '3.8'
```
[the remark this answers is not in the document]
hlolli: I'm not sure, I was following the documentation of 3.8~3.9 and not 3.0. I'm happy to revert this.

litestream.yaml
hlolli: it's been adjusted to follow the envoy pattern

docker-compose.yaml
hlolli: I need to look deeper into combining those two 👍

litestream.yaml
hlolli: no what am I saying :D I need to first push my local changes 🤦

hlolli:
I did git mistakes and I continue this here: https://github.com/ar-io/ar-io-node/pull/113

PR #113, opened 2024-04-30
feat(litestream): initialise

hlolli:


litestream/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM litestream/litestream:0.3.13
```
dtfiedler, line 1: this image does not have `ytt` or curl

you can do something like this to load `ytt` into the container bin and use it to map env values at run time

https://github.com/ar-io/arns-service/blob/litestream/litestream/Dockerfile

litestream/litestream.template.yaml
```diff
@@ -0,0 +1,12 @@
...
+  - path: /app/data/sqlite/core.db
+    replicas:
+      - type: s3
+        endpoint: #@ data.values.AR_IO_SQLITE_BACKUP_S3_BUCKET_URL
+        name: ar_io_node_core.backup.db
+        bucket: #@ data.values.AR_IO_SQLITE_BACKUP_S3_BUCKET_NAME
+        region: #@ data.values.AR_IO_SQLITE_BACKUP_S3_BUCKET_REGION
+        forcePathStyle: true
+        sync-interval: 1s
+        access-key-id: #@ data.values.AR_IO_SQLITE_BACKUP_S3_BUCKET_ACCESS_KEY
```
djwhitt, line 11: I'd ideally like the ability to use the EC2 instance role here if no explicit keys are used. That probably the default if the values aren't set explicitly. Putting a conditional around this part of the template based on whether `data.values.AR_IO_SQLITE_BACKUP_S3_BUCKET_ACCESS_KEY` is set should accomplish that.

docker-compose.yaml
```diff
@@ -1,5 +1,72 @@
----
-version: '3.0'
+version: '3.8'
```
djwhitt, line 1: As long as these changes still work, let's leave this at `3.0`.

docker-compose.yaml
```diff
@@ -1,5 +1,72 @@
...
+
+volumes:
```
djwhitt, line 3: I like this `volumes` section. Did you do any testing with it to confirm bind mounts still work as expected?

litestream/litestream.template.yaml
djwhitt: I also wonder if these can be set in such a way that they'd apply to multiple replicas when we add them.

djwhitt (commented):
This is looking pretty good. Have you gotten it to successfully write to S3?

litestream/litestream.template.yaml
dtfiedler: litestream will use the instance creds by default - you do not need to set these

djwhitt: Nice. I think we do still want support for setting them though. There are probably some users who will run on their own hardware but still backup to S3.

docker-compose.yaml
hlolli: sure, it seems furthermore that version is itself obsolete

```
$ docker compose --profile litestream up
WARN[0000] /home/ubuntu/ar-io-node/docker-compose.yaml: `version` is obsolete
```

litestream/Dockerfile
hlolli: fixed

docker-compose.yaml
hlolli: it's now been tested, and it works after making sure all mountpoints exist

litestream/litestream.template.yaml
djwhitt: The new conditional looks good. I think we should still probably do this though:
> I also wonder if these can be set in such a way that they'd apply to multiple replicas when we add them.
In other words, we should make those global settings.

hlolli: good idea, will implement

hlolli: done and I verified that it's still working with these keys in top-level scope

djwhitt (commented):
This looks good. I'm going to go ahead and contradict the original ticket a bit and say let's go ahead and add all the DBs (`core.db`, `bundles.db`, `data.db`, `moderation.db`) to the config since it seems easy. I was originally thinking we might need a separate service for each DB.

docker-compose.yaml
```diff
@@ -1,6 +1,74 @@
...
+    driver_opts:
+      type: none
+      o: bind
+      device: ${WARP_CACHE_PATH:-./data/warp/cache}
+  wallets-data:
+    driver: local
+    driver_opts:
+      type: none
+      o: bind
+      device: ${WALLETS_PATH:-./wallets}
```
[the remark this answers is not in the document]
hlolli: fixed

PR #114, opened 2024-04-30
feat(gql): pending txs and data items available via gql

karlprieb:


src/database/standalone-sqlite.ts
```diff
@@ -1565,8 +1571,9 @@ export class StandaloneSqliteDatabaseWorker {
     if (bundledIn === undefined || Array.isArray(bundledIn)) {
       sqlParts.push(`SELECT * FROM (${itemsFinalSql})`);
     }
+
     sqlParts.push(
-      `ORDER BY 1 ${sqlSortOrder}, 2 ${sqlSortOrder}, 3 ${sqlSortOrder}`,
+      `ORDER BY 1 ${sqlSortOrder}, 2 ${sqlSortOrder}, 3 ${sqlSortOrder}, indexed_at ${sqlSortOrder}`,
```
djwhitt, line 1576: Let's change this to `4` instead of `indexed_at`. We don't need the field name for performance here, and I don't want us wondering why this is an exception to the pattern in the future.

djwhitt (commented):
One minor comment about he outer `ORDER BY`.

src/database/standalone-sqlite.ts
djwhitt: As @karlprieb pointed out on Slack, this also requires shuffling the field order in the inner queries.

karlprieb: Done here [`f10117e` (#114)](https://github.com/ar-io/ar-io-node/pull/114/commits/f10117efd6371479afc4c6ed7c568382533a1993)

djwhitt (approved):

PR #119, opened 2024-05-17
feat(composite-client): try to get pending txs from peers first

karlprieb:


src/arweave/composite-client.ts
```diff
@@ -415,24 +415,14 @@ export class ArweaveCompositeClient
...
-              url: `/unconfirmed_tx/${txId}`,
-            })
-            .then((response) => {
-              return response.data;
-            });
-        }
-
         return this.peerGetTx(txId)
           .catch(async () => {
+            const url = `/${isPendingTx ? 'unconfirmed_tx' : 'tx'}/${txId}`;
```
djwhitt, line 420: We need this same logic `peerGetTx`: https://github.com/ar-io/ar-io-node/blob/ec9f2f038d4c67f74e2289d1a46210142179252d/src/arweave/composite-client.ts#L379

djwhitt (changes requested):
This simplifies the request from the trusted node, but it still needs an adjustment to fetch from peers.

src/arweave/composite-client.ts
```diff
@@ -415,24 +415,14 @@ export class ArweaveCompositeClient
...
+
             // Request TX from trusted node if peer fetch failed
             return this.trustedNodeRequestQueue.push({
               method: 'GET',
-              url: `/tx/${txId}`,
+              url,
```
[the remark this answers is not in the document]
djwhitt: Not bad. This is actually a reasonable suggestion.

src/arweave/composite-client.ts
karlprieb: Fixed here [`f01b214` (#119)](https://github.com/ar-io/ar-io-node/pull/119/commits/f01b214c2d361e2de70b806def6216b55adf88e5)

djwhitt (approved):

PR #121, opened 2024-05-20
feat(metrics): add arweave_tx_fetch_total metric

karlprieb:


src/data/gateway-data-source.ts
```diff
@@ -86,6 +87,8 @@ export class GatewayDataSource implements ContiguousDataSource {
       responseType: 'stream',
     });
 
+    metrics.arweaveTxFetchCounter.inc({ node_type: 'trusted' });
```
djwhitt, line 90: This one should be removed since it's a data rather than header request.

djwhitt (changes requested):

src/data/gateway-data-source.ts
karlprieb: Done

karlprieb:
@coderabbitai review

djwhitt (approved):

PR #134, opened 2024-05-29
feat(localstack): use persisted state by default PE-6154

fedellen:
This PR updates the docker compose to take advantage of persisted localstack state between docker compose. This way, the sidecar bundler pipeline can continue to access data items and bundles that may be in flight during restarts

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -237,8 +237,10 @@ services:
       - NODE_ENV=${NODE_ENV:-local}
       - ARWEAVE_WALLET=${ARWEAVE_WALLET:-$BUNDLER_ARWEAVE_WALLET}
       - TURBO_OPTICAL_KEY=${TURBO_OPTICAL_KEY:-$BUNDLER_ARWEAVE_WALLET}
+      - PERSIST_DEFAULT=${PERSIST_LOCALSTACK:-1} # set to 0 to disable persistence of resources and s3 objects between restarts
     volumes:
       - '${LOCALSTACK_VOLUME_DIR:-./volume}:/var/lib/localstack'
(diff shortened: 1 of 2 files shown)


docker-compose.yaml
```diff
@@ -237,8 +237,10 @@ services:
       - NODE_ENV=${NODE_ENV:-local}
       - ARWEAVE_WALLET=${ARWEAVE_WALLET:-$BUNDLER_ARWEAVE_WALLET}
       - TURBO_OPTICAL_KEY=${TURBO_OPTICAL_KEY:-$BUNDLER_ARWEAVE_WALLET}
+      - PERSIST_DEFAULT=${PERSIST_LOCALSTACK:-1} # set to 0 to disable persistence of resources and s3 objects between restarts
     volumes:
       - '${LOCALSTACK_VOLUME_DIR:-./volume}:/var/lib/localstack'
+      - '${LOCALSTACK_VOLUME_DIR:-./volume}:/persisted-data'
```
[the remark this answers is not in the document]
fedellen: the volume instructions come from here: https://github.com/GREsau/localstack-persist?tab=readme-ov-file#configuration

fwiw I've only tested this on macOS

image

djwhitt, line 243: Let's use `./data/localstack` instead of `./volume` for localstack data.

djwhitt (changes requested):

docker-compose.yaml
fedellen: 7b5ebba changed here. and removed the container name so it will open now as `ar-io-node-localstack-1`

djwhitt (approved):

PR #135, opened 2024-05-30
ci(workflows): also run core tests on macos-latest

hlolli:


dtfiedler (approved):
👏

dtfiedler:
i was premature - looks like they are failing

PR #136, opened 2024-05-31
feat(queue data item): init admin route for accepting data item headers before they are bundled PE-6220

fedellen:
This PR adds a queue data item admin endpoint 

would still like to add E2E tests and tests around upsert

manual test example:

```sh

curl -X POST http://localhost:3000/ar-io/admin/queue-data-item \
  -H "Authorization: Bearer secret" \
  -H "Content-Type: application/json" \
  -d '[{
      "id": "cTbz16hHhGW4HF-uMJ5u8RoCg9atYmyMFWGd-kzhF_Q",
      "owner": "owner",
      "owner_address": "owner_address",
      "signature": "signature",
      "target": "target",
      "anchor": "anchor",
      "content_type": "content_type",
      "data_size": 1234,
      "tags": [
          { "name": "QnVuZGxlLUZvcm1hdA", "value": "YmluYXJ5" },
          { "name": "QnVuZGxlLVZlcnNpb24", "value": "Mi4wLjA" }
      ]
  }]'

```

### DIFF
--- a/src/database/standalone-sqlite.ts
+++ b/src/database/standalone-sqlite.ts
@@ -526,6 +526,12 @@ export class StandaloneSqliteDatabaseWorker {
 
         this.stmts.bundles.upsertNewDataItem.run({
           ...rows.newDataItem,
+          root_id: item.root_tx_id,
+          parent_id: item.parent_id,
+          index: item.index,
+          parent_index: item.parent_index,
+          data_hash: item.data_hash,
+          data_offset: item.data_offset,
           height,
         });
       },
--- a/src/database/standalone-sqlite.ts
+++ b/src/database/standalone-sqlite.ts
@@ -309,7 +309,7 @@ export function dataItemToDbRows(item: NormalizedDataItem, height?: number) {
       target: fromB64Url(item.target),
       data_offset: item.data_offset,
       data_size: item.data_size,
-      content_type: contentType,
+      content_type: contentType ?? item.content_type,
       tag_count: item.tags.length,
       indexed_at: currentUnixTimestamp(),
     },
--- a/src/database/standalone-sqlite.ts
+++ b/src/database/standalone-sqlite.ts
@@ -526,12 +526,6 @@ export class StandaloneSqliteDatabaseWorker {
 
         this.stmts.bundles.upsertNewDataItem.run({
           ...rows.newDataItem,
-          root_id: item.root_tx_id,
-          parent_id: item.parent_id,
-          index: item.index,
-          parent_index: item.parent_index,
-          data_hash: item.data_hash,
-          data_offset: item.data_offset,
           height,
         });
       },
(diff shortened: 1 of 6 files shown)


src/routes/ar-io.ts
```diff
@@ -128,3 +128,83 @@ arIoRouter.post(
...
+        dataItemHeaders.length === 0
+      ) {
+        res.status(400).send('Must provide array of data item headers');
+        return;
+      }
+
+      for (const dataItemHeader of dataItemHeaders) {
+        const {
+          id,
+          // bundlr_signature, // not user, we have admin key
```
fedellen, line 173: can remove, we won't check. just admin use key

src/database/standalone-sqlite.ts
```diff
@@ -526,6 +526,12 @@ export class StandaloneSqliteDatabaseWorker {
 
         this.stmts.bundles.upsertNewDataItem.run({
           ...rows.newDataItem,
+          root_id: item.root_tx_id,
```
djwhitt, line 529: Overriding these shouldn't be necessary. The call to `dataItemToDbRows` should already be setting them.

src/types.d.ts
```diff
@@ -529,3 +529,16 @@ export type KVBufferStore = {
   del(key: string): Promise<void>;
   has(key: string): Promise<boolean>;
 };
+
+/** Accepted in queue data item route fields as normalized b64 */
+export interface QueueDataItemHeaders {
```
djwhitt, line 534: This type could be moved into `src/routes/ar-io.ts` since it's not part of a public interface.

djwhitt (commented):
Nice job digging into the code base! I commented on a couple minor issues. You will also need to update this query to make sure parent_id is set correctly when the bundle is imported: https://github.com/ar-io/ar-io-node/blob/develop/src/database/sql/bundles/import.sql#L100

src/database/standalone-sqlite.ts
```diff
@@ -309,7 +309,7 @@ export function dataItemToDbRows(item: NormalizedDataItem, height?: number) {
       target: fromB64Url(item.target),
       data_offset: item.data_offset,
       data_size: item.data_size,
-      content_type: contentType,
+      content_type: contentType ?? item.content_type,
```
fedellen, line 312: this allows content type to first come from Content-Type tag if exists, otherwise use from the incoming normalized data item headers if that exists

fedellen: fwiw I found this when there was `null` in content type in my e2e testing. should we also default to application/octet-stream here?

src/database/standalone-sqlite.ts
fedellen: removed here: [50539d2](https://github.com/ar-io/ar-io-node/pull/136/commits/50539d2beea91b2e4fa9e497b8884c88660dfc5a)

src/database/standalone-sqlite.ts
djwhitt: Good question, but nah, we lose the ability to distinguish the case when that's being set explicitly vs omitted if we put a default here. We have default on the way out when we serve the data.

djwhitt (approved):

PR #137, opened 2024-06-05
 Implement Fuzzy Search Using PostgreSQL

foxoffire33:
As part of the project I’m working on, I need to implement fuzzy search functionality. For example, if a user searches for an artist and makes a typo, I still want to return the correct result. To achieve this, I plan to use a feature of the PostgreSQL database. Currently, I am testing this on PostgreSQL.

Additionally, I have noticed that sorting is currently broken, and I might need some help resolving this issue. The database structure will follow in a future update.

I would appreciate any feedback you might have on this approach.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,6 +1,4 @@
 ---
-version: '3.0'
-
 services:
   envoy:
     image: ghcr.io/ar-io/ar-io-envoy:${ENVOY_IMAGE_TAG:-latest}
--- a/migrations/2024.05.28T18.53.33.core.createViewForWildcardSearch.sql
+++ b/migrations/2024.05.28T18.53.33.core.createViewForWildcardSearch.sql
@@ -1,27 +1,7 @@
 -- up migration
-create view transactionsTagsDecoded as
-select distinct *
-from (select transaction_id,
-             tag_name_hash,
-             cast(tag_names.name as varchar)   as tag_name,
-             tag_value_hash,
-             cast(tag_values.value as varchar) as tag_value,
-             transaction_tag_index,
-             height
-      from stable_transaction_tags
-               join tag_names on tag_names.hash = tag_name_hash
-               join tag_values on tag_values.hash = tag_value_hash
-      union
-      select transaction_id,
-             tag_name_hash,
-             cast(tag_names.name as varchar)   as tag_name,
-             tag_value_hash,
-             cast(tag_values.value as varchar) as tag_value,
-             transaction_tag_index,
-             height
-      from new_transaction_tags
-               left join tag_names on tag_names.hash = tag_name_hash
-               left join tag_values on tag_values.hash = tag_value_hash);
-
-create index tagNamesIndex on tag_names (hash, name);
-create index tagValuesIndex on tag_values (hash, value);
+CREATE INDEX IF NOT EXISTS tag_names_index ON tag_names (hash, name);
+CREATE INDEX IF NOT EXISTS tag_values_index ON tag_values (hash, value);
+CREATE INDEX IF NOT EXISTS new_transaction_tags_index ON new_transaction_tags (transaction_id, tag_name_hash, tag_value_hash);
+CREATE INDEX IF NOT EXISTS stable_transaction_tags_index ON stable_transaction_tags (transaction_id, tag_name_hash, tag_value_hash);
+CREATE INDEX IF NOT EXISTS new_blocks_index ON new_blocks (height);
+CREATE INDEX IF NOT EXISTS stable_blocks_index ON stable_blocks (height);
--- a/migrations/2024.05.28T18.53.33.core.createViewForWildcardSearch.sql
+++ b/migrations/2024.05.28T18.53.33.core.createViewForWildcardSearch.sql
@@ -1,0 +1,27 @@
+-- up migration
+create view transactionsTagsDecoded as
+select distinct *
+from (select transaction_id,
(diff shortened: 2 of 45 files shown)


djwhitt:
@foxoffire33 I've been thinking about this more today. What I'm initially curious about is what kind of fuzzy search you want to provide and which specific Postgres features you intend to use for it. (apologies if this is visible in the diffs. I probably won't be able to review those in detail till next week) I'm also curious if you considered using the webhook support to populate some external index (e.g., OpenSearch or similar), and if not, why.

Regarding potential integration - how much effort are you willing to put into maintaining this? We would like to add Postgres support in `develop`/`main` at some point, but envisioned using a somewhat different schema from SQLite when we eventually implemented it. If you're interested in pursuing that, I could give you some info about the schema design we had in mind.

foxoffire33:
I tried to do it with Meilisearch first, but it doesn't fit the case perfectly. Implementing the functionality I need would be similar in cost to changing the gateway itself.

Since this is for a project I'm working on, I'm happy to maintain a Postgres solution for fuzzy search as long as we need it.

I'm also interested in the database design you already have in place, and I might be able to contribute some code for that.

djwhitt:
> I tried to do it with Meilisearch first, but it doesn't fit the case perfectly. Implementing the functionality I need would be similar in cost to changing the gateway itself.

Gotcha, just wanted to make sure you were aware of the webhook option.

> Since this is for a project I'm working on, I'm happy to maintain a Postgres solution for fuzzy search as long as we need it.

Great! What's the current status of the PR? Is it working for your use case now or are you still sorting through issues?

> I'm also interested in the database design you already have in place, and I might be able to contribute some code for that.

Also great! I don't think I'll get to it till next week at this point, but I'll write up some details soon.

foxoffire33:
I'm still working on it. Do not have a lot of time.

 There is an issue with sorting in queries. I will provide an update when I know more.

djwhitt:
Long overdue update from our side - we're working on a version of the schema I'd recommend on an internal legacy database (different code base). It's similar to the SQLite schema but uses a single height partitioned table in Postgres as opposed to separate 'new_*' and 'stable_*' tables. We represent `null` heights with a very large positive integer since partition keys can't be nullable. The advantage of this approach is that all the new/stable flushing is automatically handled by Postgres.

foxoffire33:
is it possible to share the sql schema?

djwhitt:
Sure. This should give you the general idea, but keep in mind this is from a different DB with some legacy decisions (strings instead of binary fields):

```sql
CREATE TABLE data_items (
    height integer,
    height_partition_key integer,
    id character(43) NOT NULL,
    raw_offset bigint,
    raw_size bigint,
    signature character varying, -- only here to support ANS102
    signature_type bigint,
    signature_offset bigint,
    signature_size bigint,
    owner_address character(43),
    owner_offset bigint,
    owner_size bigint,
    data_offset bigint,
    data_size bigint,
    target character(43),
    quantity character varying,
    reward character varying,
    anchor character varying,
    content_type character varying,
    format smallint,
    created_at timestamp without time zone DEFAULT now() NOT NULL,
    updated_at timestamp without time zone DEFAULT now() NOT NULL,
    parent character(43),
    parent_root_tx_offset bigint, -- offset of the parent relative to the root TX
    root_tx_id character(43), -- same as parent when it's not a nested data-item
    PRIMARY KEY (height_partition_key, id)
) PARTITION BY RANGE (height_partition_key);

CREATE INDEX data_item_id_idx ON data_items (id);
CREATE INDEX data_item_height_part_key_desc_id_idx ON data_items (height_partition_key DESC, id);
CREATE INDEX data_item_owner_address_height_part_key_id_idx ON data_items (owner_address, height_partition_key, id);
CREATE INDEX data_item_owner_address_height_part_key_desc_id_idx ON data_items (owner_address, height_partition_key DESC, id);
CREATE INDEX data_item_target_height_part_key_id_idx ON data_items (target, height_partition_key, id);
CREATE INDEX data_item_target_height_part_key_desc_id_idx ON data_items (target, height_partition_key DESC, id);

CREATE TABLE data_item_tags (
    height integer,
    height_partition_key integer,
    tx_id character(43) NOT NULL,
    index integer NOT NULL,
    name_hash bytea NOT NULL,
    value_hash bytea NOT NULL,
    created_at timestamp without time zone DEFAULT now() NOT NULL,
    updated_at timestamp without time zone DEFAULT now() NOT NULL,
    PRIMARY KEY (height_partition_key, tx_id, index)
) PARTITION BY RANGE (height_partition_key);


CREATE INDEX data_item_tags_name_value_height_part_tx_id_idx ON data_item_tags (name_hash, value_hash, height_partition_key, tx_id);
CREATE INDEX data_item_tags_name_value_height_part_key_desc_tx_id_idx ON data_item_tags (name_hash, value_hash, height_partition_key DESC, tx_id);

CREATE TABLE data_item_tag_names (
    name_hash bytea NOT NULL PRIMARY KEY,
    name character varying NOT NULL
);

CREATE INDEX data_item_tag_names_name_idx ON data_item_tag_names (name);

CREATE TABLE data_item_tag_values (
    value_hash bytea NOT NULL PRIMARY KEY,
    value character varying NOT NULL
);

CREATE INDEX data_item_tag_values_value_idx ON data_item_tag_values (value);
```

djwhitt:
Closing this one due to inactivity. I don't think we can integrate as is. I'm happy to engage again though if there's more progress or interest from others.

PR #138, opened 2024-06-05
feat(queue data item): enable bundle sidecar optical post to local queue data item admin endpoint PE-6107

fedellen:
depends on next upload service public image


docker-compose.yaml
```diff
@@ -247,8 +248,8 @@ services:
           'CMD-SHELL',
           'curl -f http://localhost:4566/_localstack/health || exit 1',
         ]
-      interval: 90s
+      interval: 10s
       timeout: 30s
-      retries: 1
+      retries: 10
```
fedellen, line 253: this saves a bunch of startup time when using a persisted localstack cache

djwhitt (approved):

djwhitt (approved):

PR #139, opened 2024-06-05
feat(s3-data-source): add S3DataSource

karlprieb:


src/config.ts
```diff
@@ -238,6 +238,19 @@ export const MEMPOOL_POLLING_INTERVAL_MS = +env.varOrDefault(
   '30000', // 30 seconds
 );
 
+//
+// AWS settings
+//
+//
+export const AWS_ACCESS_KEY_ID = env.varOrUndefined('AWS_ACCESS_KEY_ID');
```
karlprieb, line 245: @djwhitt should we add these variables to docs or wait for the featured to be tested before we "expose" it?

karlprieb:
@djwhitt It would be good to add some metrics for this. What do you think?

src/system.ts
```diff
@@ -77,6 +80,18 @@ process.on('uncaughtException', (error) => {
 
 const arweave = Arweave.init({});
 const arIO = ArIO.init({ contractTxId: config.CONTRACT_ID });
+const awsClient =
+  config.AWS_ACCESS_KEY_ID !== undefined &&
+  config.AWS_SECRET_ACCESS_KEY !== undefined &&
+  config.AWS_REGION !== undefined
+    ? await awsLite({
+        accessKeyId: config.AWS_ACCESS_KEY_ID,
```
djwhitt, line 88: We won't to be able to use instance roles, so let's allow for `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` being `undefined`.

test/end-to-end/data-sources.test.ts
```diff
@@ -0,0 +1,171 @@
...
+      );
+      assert.equal(
+        toB64Url(importedBundle.root_transaction_id),
+        'R4UyABK-I7bgzJVhsUZ3JdtvHrFYQBtJQFsZK1xNrJA',
+      );
+      assert.equal(importedBundle.data_item_count, 2);
+      assert.equal(importedBundle.import_attempt_count, 1);
+    });
+  });
+});
```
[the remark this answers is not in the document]
djwhitt: @karlprieb Just a nitpick, but this seems reasonable :point_up_2:

test/end-to-end/webhook.test.ts
```diff
@@ -16,7 +16,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 import { strict as assert } from 'node:assert';
-import { after, afterEach, before, beforeEach, describe, it } from 'node:test';
+import { after, before, beforeEach, describe, it } from 'node:test';
```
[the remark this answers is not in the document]
djwhitt: @karlprieb Same as above.

src/config.ts
djwhitt: I'd say go ahead and add them. That way we can direct any community members who want to test this to the docs.

fedellen (approved):

src/system.ts
arielmelendez: @karlprieb there's some examples of setting up AWS configs for local use in the Turbo source. Not sure if totally relevant here:

https://github.com/ardriveapp/upload-service/blob/dev/src/arch/s3ObjectStore.ts#L60-L72
and
https://github.com/ardriveapp/upload-service/blob/dev/src/arch/s3ObjectStore.ts#L164-L187

src/config.ts
```diff
@@ -238,6 +238,19 @@ export const MEMPOOL_POLLING_INTERVAL_MS = +env.varOrDefault(
...
+export const AWS_SECRET_ACCESS_KEY = env.varOrUndefined(
+  'AWS_SECRET_ACCESS_KEY',
+);
+export const AWS_REGION = env.varOrUndefined('AWS_REGION');
+export const AWS_ENDPOINT = env.varOrUndefined('AWS_ENDPOINT');
+export const AWS_S3_BUCKET = env.varOrUndefined('AWS_S3_BUCKET');
```
arielmelendez, line 251: These two don't have much flexibility if we need to use different buckets for different things. Might be ok for now though.

src/data/s3-data-source.ts
```diff
@@ -0,0 +1,115 @@
...
+    requestAttributes,
+  }: {
+    id: string;
+    requestAttributes?: RequestAttributes;
+  }): Promise<ContiguousData> {
+    const log = this.log.child({ method: 'getData' });
+    log.info('Fetching contiguous data from S3', {
+      id,
+      bucket: this.s3Bucket,
+      prefix: this.s3Prefix,
```
arielmelendez, line 63: Could it be useful to see the requestAttributes here as well? Also do we want these to be info or debug logs?

src/data/s3-data-source.ts
```diff
@@ -0,0 +1,115 @@
...
+        Bucket: this.s3Bucket,
+        Key: `${this.s3Prefix}/${id}`,
+        streamResponsePayload: true,
+      });
+
+      const requestOriginAndHopsHeaders: { [key: string]: string } = {};
+      let hops;
+      let origin;
+      if (requestAttributes !== undefined) {
+        hops = requestAttributes.hops + 1;
```
arielmelendez, line 77: How do we index hops? 1 = locally resolved? Or should it be 0?

src/config.ts
karlprieb: [`60e3b62` (#139)](https://github.com/ar-io/ar-io-node/pull/139/commits/60e3b626971f301ff800050cacca78bb62c8c506)

test/end-to-end/webhook.test.ts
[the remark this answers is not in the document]
karlprieb: [`5b33d7d` (#139)](https://github.com/ar-io/ar-io-node/pull/139/commits/5b33d7daa0333fd8a5bb8030d1c4981ffb1b08f8)

test/end-to-end/data-sources.test.ts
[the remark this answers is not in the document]
karlprieb: [`5b33d7d` (#139)](https://github.com/ar-io/ar-io-node/pull/139/commits/5b33d7daa0333fd8a5bb8030d1c4981ffb1b08f8)

src/system.ts
karlprieb: `aws-lite` throws if `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` are not defined.
We can have another ticket to use instance profiles as they support it

src/data/s3-data-source.ts
```diff
@@ -0,0 +1,115 @@
...
+    s3Prefix = '',
+  }: {
+    log: winston.Logger;
+    s3Client: AwsLiteS3;
+    s3Bucket: string;
+    s3Prefix?: string;
+  }) {
+    this.log = log.child({ class: 'S3DataSource' });
+    this.s3Client = s3Client;
+    this.s3Bucket = s3Bucket;
```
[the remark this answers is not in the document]
karlprieb: Localstack works fine with virtual hosts. It needs a domain, so the problem happens when you deploy locally without any domain.

This doesn't mean you can't have a local domain and make it work locally.

src/config.ts
karlprieb: Good point. We can implement multi bucket support in the future.

src/data/s3-data-source.ts
karlprieb: `requestAttributes` are just `hops` and `origin` which are not propagated to S3.

In this case we just increment `hops` and return the new value.

src/data/s3-data-source.ts
karlprieb: `hops` counts how many hops a server made to get the data, in this case when successfully fetched from S3 we just increment the number.

When getting data from other ario node they should respond with a new `hops` value (amount of hops they made to get the data)

src/config.ts
```diff
@@ -71,7 +71,7 @@ export const TRUSTED_GATEWAY_URL = env.varOrDefault(
 export const ON_DEMAND_RETRIEVAL_ORDER = env
   .varOrDefault(
     'ON_DEMAND_RETRIEVAL_ORDER',
-    'ario-peer,trusted-gateway,chunks,tx-data',
+    's3,ario-peer,trusted-gateway,chunks,tx-data',
```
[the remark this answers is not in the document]
karlprieb: It will only add `S3DataSource` if you instantiate the class (requires AWS variables to be set)

https://github.com/ar-io/ar-io-node/blob/5b33d7daa0333fd8a5bb8030d1c4981ffb1b08f8/src/system.ts#L304-L307

src/system.ts
djwhitt: I looked into this a bit too after adding the comment. Seems like there's no cred provider integrated that'll use instance roles by default. I think we need to retrieve keys and then switch to the instance role, but that's more work than we want to do in this ticket.

src/data/s3-data-source.ts
djwhitt: We're always counting the inbound request as a hop and adding to it. See: https://github.com/ar-io/ar-io-node/blob/develop/src/data/gateway-data-source.ts#L76

djwhitt (approved):

PR #148, opened 2024-06-12
feat(data-sources): add getData metrics

karlprieb:
Added getDataErrorsTotal, getDataStreamErrorsTotal and getDataStreamSuccesssTotal metrics for ContiguousDataSources.

### DIFF
--- a/src/arweave/composite-client.ts
+++ b/src/arweave/composite-client.ts
@@ -635,29 +635,51 @@ export class ArweaveCompositeClient
   async getData({ id }: { id: string }): Promise<ContiguousData> {
     this.failureSimulator.maybeFail();
 
-    const [dataResponse, dataSizeResponse] = await Promise.all([
-      this.trustedNodeRequestQueue.push({
-        method: 'GET',
-        url: `/tx/${id}/data`,
-      }),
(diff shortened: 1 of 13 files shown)


src/arweave/composite-client.ts
```diff
@@ -635,29 +635,51 @@ export class ArweaveCompositeClient
...
-      cached: false,
-    };
+      const size = +dataSizeResponse.data;
+      const txData = fromB64Url(dataResponse.data);
+
+      const stream = Readable.from(txData);
+
+      stream.on('error', () => {
+        metrics.getDataStreamErrorsTotal.inc({
+          class: 'ArweaveCompositeClient',
```
djwhitt, line 661: I *think* we can use `this.constructor.name` to get the class name (might have to be outside the callback).

djwhitt (commented):
Looks good so far, just added one comment about a potential way to simplify class name retrieval.

src/arweave/composite-client.ts
karlprieb: Great suggestion!

djwhitt (approved):

PR #150, opened 2024-06-14
refactor(sqlite): use null parent and root TX ID for optimistic data items PE-6284

djwhitt:
For expedience, the initial optimistic indexing code used placeholder values for root TX ID and parent. While this mostly works, it's confusing and produces inaccurate GQL results without extra post processing (optimistic data item parents are wrong). This change modifies the DB to allow null parent, root transaction ids, and offsets. It also adjusts types to allow for this and introduces extra conditionals to ensure that bundle imports and data indexing for optimistically indexed data items are not attempted. This prevents errant optimistic data items from making their way into the stable indexes.

### DIFF
--- a/migrations/2024.06.13T13.52.01.bundles.nullable-parent-and-root.sql
+++ b/migrations/2024.06.13T13.52.01.bundles.nullable-parent-and-root.sql
@@ -1,0 +1,47 @@
+DROP TABLE new_data_items;
+
+CREATE TABLE new_data_items (
+  -- Identity
+  id BLOB NOT NULL,
+  parent_id BLOB,
+  root_transaction_id BLOB,
+  height INTEGER,
+  signature BLOB NOT NULL,
+  anchor BLOB NOT NULL,
+
+  -- Ownership
+  owner_address BLOB NOT NULL,
+  target BLOB,
+
+  -- Data
+  data_offset INTEGER,
+  data_size INTEGER NOT NULL,
+  content_type TEXT,
+
+  -- Metadata
+  tag_count INTEGER NOT NULL,
+  indexed_at INTEGER NOT NULL,
+  PRIMARY KEY (id)
+);
+
+CREATE INDEX new_data_items_parent_id_id_idx ON new_data_items (parent_id, id);
+CREATE INDEX new_data_items_root_transaction_id_id_idx ON new_data_items (root_transaction_id, id);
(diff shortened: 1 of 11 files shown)


migrations/2024.06.13T13.52.01.bundles.nullable-parent-and-root.sql
```diff
@@ -0,0 +1,47 @@
+DROP TABLE new_data_items;
+
+CREATE TABLE new_data_items (
```
fedellen, line 3: will this effectively reset each node's `new_data_items` when the migration performs?

src/workers/ans104-data-indexer.ts
```diff
@@ -91,23 +91,53 @@ export class Ans104DataIndexer {
...
+          }
+        } else {
+          log.warn('Skipping data item data indexing due to missing hash.');
+        }
+
+        // Index data offset and size for ID to parent ID relationship
+        if (
+          item.parent_id != null &&
+          item.data_offset != null &&
+          item.data_size != null
```
fedellen, line 128: nit: since its only one type or the other, I think we could simplify these null checks with a broader `isOptimisiticDataItem` type guard that has a type like `(dataItem: NormalizedDataItem): dataItem is NormalizedOptimisticDataItem => dataItem.parent_id != null`

fedellen (approved):

migrations/2024.06.13T13.52.01.bundles.nullable-parent-and-root.sql
karlprieb: Yes

karlprieb (approved):

migrations/2024.06.13T13.52.01.bundles.nullable-parent-and-root.sql
djwhitt: Yeah, it does, but it shouldn't be a big deal. They'll just pick up and start importing again at the stable threshold. Also we have bundle import retry mechanisms that should kick in if anything is missed.

src/workers/ans104-data-indexer.ts
djwhitt: I like the general suggestion, but in this particular case, I'm paranoid about things slipping through with an improper type. I also like that I can look at the list of things being checked here and match them up with the DB calls. That said, I would probably go with the type guard if these conditionals weren't around DB inserts or our I was more confident about our typing overall.

PR #152, opened 2024-06-19
feat(manifests): add support for fallback id

karlprieb:
I'm still working on e2e tests


src/lib/encoding.ts
```diff
@@ -244,6 +249,16 @@ export function parseManifestStream(stream: Readable): EventEmitter {
   });
 
   pipeline.on('end', () => {
+    if (fallbackData !== undefined) {
```
karlprieb, line 252: To ensure that we only handle fallback at the end

src/lib/encoding.ts
```diff
@@ -305,12 +341,10 @@ export function parseManifestStream(stream: Readable): EventEmitter {
     ) {
       pathCount++;
       const p = keyPath[1];
+      paths[p] = data;
```
karlprieb, line 344: @djwhitt what was the reason to empty the paths object before?
I have to keep it to handle fallback paths at the end of the stream.

src/routes/data/handlers.ts
```diff
@@ -348,7 +347,8 @@ const sendManifestResponse = async ({
     // Add a trailing slash if needed
     if (req.path === `/${id}`) {
       // Extract query string using the url module
-      const queryString = url.parse(req.url).search ?? '';
+      const queryString =
```
karlprieb, line 350: It was giving me deprecation warnings when using `url.parse` to target an param

src/lib/encoding.ts
djwhitt: > I have to keep it to handle fallback paths at the end of the stream.

Main reason to clear it out is to conserve memory because manifests are unbounded in size. We should be able to continue discarding it as soon as we have seen both index and fallback paths I think.

karlprieb: As `fallback` is limited to `id` the original `paths` behavior was kept.

src/lib/encoding.test.ts
```diff
@@ -245,39 +246,78 @@ describe('Transaction message pack encoding and decoding functions', () => {
...
+        assert.equal(id3, '0543SMRGYuGKTaqLzmpOyK4AxAB96Fra2guHzYxjRGo');
+      });
+
+      it('should return fallback if the path is not found', async () => {
+        const id = await resolveManifestStreamPath(
+          exampleManifestStreamV020(),
+          'missing',
+        );
+        assert.equal(id, 'cG7Hdi_iTQPoEYgQJFqJ8NMpN4KoZ-vH_j7pG4iP7NI');
+      });
```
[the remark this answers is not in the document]
karlprieb: Good suggestion!

djwhitt (approved):

PR #153, opened 2024-06-20
feat(manifests): add support for index id

karlprieb:


src/lib/encoding.ts
```diff
@@ -329,9 +360,11 @@ export function parseManifestStream(stream: Readable): EventEmitter {
       pathCount++;
       const p = keyPath[1];
       emitter.emit('path', { path: p, id: data });
-      if (indexPath === undefined) {
-        paths[p] = data; // Maintain map of paths for use later
-      } else if (p === indexPath) {
+      if (indexProps.path === undefined) {
+        if (indexProps.id !== undefined) {
```
djwhitt, line 364: Is this condition correct? Don't we want the reverse? (only keep paths if there's no index ID specified) We should also probably add unit tests for these cases.

src/lib/encoding.ts
```diff
@@ -295,18 +297,47 @@ export function parseManifestStream(stream: Readable): EventEmitter {
...
+      currentKey === 'id'
+    ) {
+      indexProps.id = data;
+      emitter.emit('index', {
+        id: data,
+      });
+      paths = {};
+    }
+
+    // Fallback - { "fallback": { "fallback": { "id": "<data-id>" } }
```
djwhitt, line 333: Comment typo here - double nested `fallback`.

src/lib/encoding.ts
```diff
@@ -295,18 +297,47 @@ export function parseManifestStream(stream: Readable): EventEmitter {
...
+    if (
+      keyPath.length === 1 &&
+      keyPath[0] === 'fallback' &&
+      isManifestV2 &&
+      currentKey === 'id'
+    ) {
+      fallbackId = data;
     }
 
     // Fallback - { "fallback": { "fallback": { "id": "<data-id>" } }
```
djwhitt, line 343: Same typo as above.

src/lib/encoding.ts
```diff
@@ -295,18 +297,47 @@ export function parseManifestStream(stream: Readable): EventEmitter {
...
       keyPath[0] === 'index' &&
       currentKey === 'path'
     ) {
-      indexPath = data;
+      indexProps.path = data;
       // Resolve if the path id is already known
-      if (indexPath !== undefined && paths[indexPath] !== undefined) {
-        emitter.emit('index', { path: indexPath, id: paths[indexPath] });
-        paths = {};
+      if (
```
djwhitt, line 308: Let's add an `indexProps.id === undefined` condition here.

djwhitt (changes requested):
I think there are a few edge cases we still need to handle.

src/lib/encoding.ts
karlprieb: In the case someone sets `path` and `id`?

djwhitt: Exactly. We want the `id` to win in that case (same as legacy gateway).

karlprieb: Adding a `id` check will not prevent `path` to win. It will only win if `id` is parsed first.
If we want to prioritize `id` we need to ensure check if `id` exists in the case of `path` being parsed first. With this check it will only check if `id` was already parsed or not (we don't know if it exists later at this point when `path` is parsed first than `id`).

djwhitt: Hmm... are you thinking about the case where `index` is after `paths`? It would prevent it otherwise, correct?

karlprieb: Yes

djwhitt (approved):
Nice. Looks good!

PR #155, opened 2024-06-25
feat(indexing): add signature_type to data items

karlprieb:


djwhitt (changes requested):
Issue in the down migration. Otherwise this looks good.

djwhitt (approved):

PR #160, opened 2024-06-27
chore(env): create env function to generate random variable

dtfiedler:
Replaces the current logic with a function that we can use again if necessary


karlprieb:
Is this still a draft? Looks good to me

karlprieb (approved):

djwhitt (approved):

PR #161, opened 2024-07-02
feat(data-handlers): accept multiple range request

karlprieb:


src/routes/data/handlers.ts
```diff
@@ -192,8 +192,93 @@ const handleRangeRequest = (
...
+    res.setHeader('Content-Type', `multipart/byteranges; boundary=${boundary}`);
+    res.setHeader('Accept-Ranges', 'bytes');
+
+    if (req.method === REQUEST_METHOD_HEAD) {
+      res.end();
+      data.stream.destroy();
+      return;
+    }
+
+    const rangeStreams: { range: any; stream: PassThrough }[] = [];
```
hlolli, line 217: nit-pick: could be this type? https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/range-parser/index.d.ts#L18-L21

karlprieb: Yes

karlprieb:
@hlolli and I were discussing in Slack if we should add a range limit and also if we should combine ranges (0-0, 1-1 -> 0-1).

Limiting the range can help us block some kids from abusing the server, but nothing holds people to trigger several requests to abuse it.

Currently, we combine ranges, but I don't like the idea. If a client wants to parse some bytes and is counting on doing this in the order of the chunks it receives, it will need to develop an additional parsing logic to get only the bytes they requested.

What do you @ar-io/core-team think about it?

dtfiedler (approved):
@karlprieb - it looks like `rangeParser` does most of the work ensuring the ranges provided are valid before any streams are created. i'd double check there are no known issues on that library atm, but otherwise LG.

as for the range limit - eh, indifferent. I'm not sure what a reasonable limit would be. r.e. `combine` the [default behavior](https://www.npmjs.com/package/range-parser#combine) is `false` - was there any specific reason set on the original commit for setting it to `true`? if so, i'd keep it.

nice work!

src/routes/data/handlers.test.ts
```diff
@@ -163,8 +168,28 @@ describe('Data routes', () => {
         });
     });
 
-    // Multiple ranges are not yet supported
-    it('should return 416 status code for a range request with multiple ranges', async () => {
+    it('should return 416 status code for a unsatisfiable range request', async () => {
```
arielmelendez, line 171: Perhaps this should be two separate tests:
- one is an unsatisfiable range because the tx doesn't exist
- the other is an unsatisfiable range because it's beyond the bounds of the tx size (this could actually be 2 tests - one where just the end of the range is invalid and one where the start AND the end of the range are invalid)

src/routes/data/handlers.test.ts
```diff
@@ -178,13 +203,109 @@ describe('Data routes', () => {
...
+          const boundary = res.boundary;
+          assert.equal(
+            res.get('Content-Type'),
+            `multipart/byteranges; boundary=${boundary}`,
+          );
+
+          // binary response data is in res.body as a buffer
+          assert.ok(Buffer.isBuffer(res.body));
+
+          // As  the server respond with a \r\n linebreak and JS template literals use \n linebreaks, explicit \r\n is needed in the expected response
```
arielmelendez, line 228: Thanks for the explanation here

arielmelendez: A thought - maybe a multiline string followed by `.split('\n').join('\r\n')` would be more readable?

src/routes/data/handlers.test.ts
```diff
@@ -178,13 +203,109 @@ describe('Data routes', () => {
...
+            `--${boundary}\r\n` +
+            `Content-Type: application/octet-stream\r\n` +
+            `Content-Range: bytes 4-5/10\r\n` +
+            `\r\n` +
+            'in\r\n' + // second range string
+            `--${boundary}--\r\n`;
+          assert.equal(res.body.toString(), expectedResponse);
+        });
+
+      const getInvertedRange = request(app)
```
arielmelendez, line 244: Might be preferable to break each of these out as their own test case with helpful describe block. Would give more granular feedback on test failures.

karlprieb:
>  r.e. `combine` the [default behavior](https://www.npmjs.com/package/range-parser#combine) is `false` - was there any specific reason set on the original commit for setting it to `true`? if so, i'd keep it.
> 
> nice work!

@dtfiedler It was already set to `true` so I just kept it, but I'll change it to false :)

src/routes/data/handlers.test.ts
karlprieb: Good suggestion. I believe it will work fine and will make it way more readable :)

src/routes/data/handlers.test.ts
karlprieb: When tx doesn't exist, we respond with a 404. The check happens before `handleRangeRequest` get into action.

hlolli (approved):

src/routes/data/handlers.ts
karlprieb: fixed here [`ff9f4d1` (#161)](https://github.com/ar-io/ar-io-node/pull/161/commits/ff9f4d11d287d46c0dbcd5d54ac84ab72ca923a7)

src/routes/data/handlers.test.ts
karlprieb: fixed here [`51cb78b` (#161)](https://github.com/ar-io/ar-io-node/pull/161/commits/51cb78bf59e439b97422183521d7acf32872dd96)

src/routes/data/handlers.test.ts
karlprieb: I used `replace` instead
[`51cb78b` (#161)](https://github.com/ar-io/ar-io-node/pull/161/commits/51cb78bf59e439b97422183521d7acf32872dd96)

arielmelendez (approved):

src/routes/data/handlers.test.ts
```diff
@@ -225,23 +220,37 @@ describe('Data routes', () => {
...
+Content-Range: bytes 1-2/10
+
+es
+--${boundary}
+Content-Type: application/octet-stream
+Content-Range: bytes 4-5/10
+
+in
+--${boundary}--
+`.replace(/\n/g, '\r\n');
```
dtfiedler, line 235: this always makes me sad

src/routes/data/handlers.ts
```diff
@@ -117,9 +117,7 @@ const handleRangeRequest = (
   data: ContiguousData,
   dataAttributes: ContiguousDataAttributes | undefined,
 ) => {
-  const ranges = rangeParser(data.size, rangeHeader, {
-    combine: true,
-  });
+  const ranges = rangeParser(data.size, rangeHeader);
```
dtfiedler, line 120: just a thought - this may be something worth noting in the CHANGELOG (i.e. "range requests with adjacent boundaries will no longer be combined and returned as a single stream"). just so we have something to point to if it does introduce unintentional issues with certain HTTP clients

dtfiedler (approved):

PR #164, opened 2024-07-11
feat(bundles): add new data item properties

karlprieb:
Added new columns:
 - offset
 - size
 - owner_offset
 - owner_size
 - signature_offset
 - signature_size

`processBundleStream` (previously `processStream`) was added. This function was extended from arbundles to expose more properties. Also a small performance improvement on how we handle buffers was done. Original `processStream`: https://github.com/Irys-xyz/arbundles/blob/f8b5834803307a9308b028d37c6fe3975365e661/src/stream/index.ts

`AVSCTap` (from arbundles) was added. This class was rewritten based on thesame class from arbundles to ensure better performance and readability while handling data item tags.
Original `AVSCTap`: https://github.com/Irys-xyz/arbundles/blob/44a56130786c52715dec7a764b93217e08bccfbf/src/tags.ts


src/lib/bundles/bundles.ts
```diff
@@ -0,0 +1,308 @@
...
+      ownerOffset,
+      ownerSize: pubLength,
+      sigName,
+      signature: Buffer.from(signature).toString('base64url'),
+      signatureOffset,
+      signatureSize: sigLength,
+      signatureType: signatureType,
+      size: length,
+      tags,
+      target: Buffer.from(target).toString('base64url'),
```
fedellen, line 279: nit: we have unit tested implementations for these b64 conversions in `lib/encoding.ts`

package.json
```diff
@@ -12,7 +12,7 @@
     "@aws-lite/client": "^0.21.7",
     "@aws-lite/s3": "^0.1.21",
     "apollo-server-express": "^3.13.0",
-    "arbundles": "^0.6.19",
+    "arbundles": "^0.11.1",
```
fedellen, line 15: should we be able to move this to a dev dependency now?

src/database/sql/bundles/import.sql
```diff
@@ -89,11 +89,13 @@ UPDATE SET
 INSERT INTO new_data_items (
   id, parent_id, root_transaction_id, height, signature, anchor,
   owner_address, target, data_offset, data_size, content_type,
-  tag_count, indexed_at, signature_type
+  tag_count, indexed_at, signature_type, offset, size, owner_offset,
+  owner_size, signature_offset, signature_size
```
fedellen, line 93: we use the `size` namespace quite a lot here (`data_size`, `owner_size`, etc) and now adding `size` itself. I understand its the total size of the data item with its headers. wondering if theres opportunity here for describing that in the column

src/lib/ans-104.ts
```diff
@@ -39,10 +38,7 @@ import {
   NormalizedDataItem,
 } from '../types.js';
 import { fromB64Url, sha256B64Url, utf8ToB64Url } from './encoding.js';
-
-/* eslint-disable */
-// @ts-ignore
-const { default: processStream } = arbundles;
+import { processBundleStream, DataItemInfo } from './bundles/bundles.js';
```
fedellen, line 41: 👋

src/lib/bundles/bundles.ts
```diff
@@ -0,0 +1,308 @@
...
+  signature: string;
+  signatureOffset: number;
+  signatureSize: number;
+  signatureType: number;
+  size: number;
+  tags: { name: string; value: string }[];
+  target: string;
+}
+
+export const processBundleStream = async (
```
djwhitt, line 115: It'll take a little extra time, but given we'll like extract this into a library soon, I think we go ahead and add unit tests for it.

djwhitt (commented):
Nice work on this. I added a few comments, but nothing critical. I'm going to go ahead and start testing this branch on PiKNiK. Also appreciate extra :eyes: on this from @arielmelendez if he has time.

src/lib/bundles/bundles.ts
karlprieb: I'll take a look. Thanks

src/database/sql/bundles/import.sql
karlprieb: @djwhitt what do you think about this?

src/lib/bundles/tags.ts
```diff
@@ -0,0 +1,127 @@
...
+      this.writeString(tag.value);
+    }
+    this.writeLong(0);
+  }
+
+  public toBuffer(): Buffer {
+    return this.buf.subarray(0, this.pos);
+  }
+
+  protected writeLong(n: number): void {
```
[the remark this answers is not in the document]
karlprieb: I couldn't find any case where it was useful.
I'll use their implementation for now while we don't have extensive tests for it.
Better be safe than sorry.

src/lib/bundles/tags.ts
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  protected writeString(s: string): void {
+    const len = Buffer.byteLength(s);
+    this.writeLong(len);
+    this.buf.write(s, this.pos, len, 'utf8');
+    this.pos += len;
+  }
+
+  protected readLong(): number {
```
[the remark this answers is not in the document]
karlprieb: See https://github.com/ar-io/ar-io-node/pull/164#discussion_r1674639783

src/lib/bundles/bundles.ts
karlprieb: I agree 👍

src/database/sql/bundles/import.sql
djwhitt: It breaks symmetry with `offset` if we only add a prefix to `size`, but we could consider `raw_offset` and `raw_size`.

src/lib/bundles/tags.ts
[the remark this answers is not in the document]
djwhitt: Gotcha. Yeah, probably best to be safe and do the same. A comment explaining we're uncertain about the utility of it would be good too.

src/lib/bundles.test.ts
```diff
@@ -0,0 +1,223 @@
...
+
+      it('when the reader provides exactly enough data', async () => {
+        const initialBuffer = Buffer.from('a');
+        const additionalBuffers = [Buffer.from('bc'), Buffer.from('def')];
+        const reader = generateBuffers(additionalBuffers);
+        const result = await readBytes(reader, initialBuffer, 6);
+        assert.deepEqual(result, Buffer.from('abcdef'));
+      });
+    });
+  });
```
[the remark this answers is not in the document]
karlprieb: Sure, what your suggestions?

src/lib/bundles.ts
```diff
@@ -0,0 +1,259 @@
...
+  return Object.values(SignatureConfig).includes(value);
+};
+
+export const getSignatureMeta = (signatureType: number): SignatureMeta => {
+  if (isValidSignatureConfig(signatureType)) {
+    return SIG_CONFIG[signatureType];
+  } else {
+    throw new Error('Invalid signature type');
+  }
+};
```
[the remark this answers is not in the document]
karlprieb: This suggestion doesn't apply to this function

djwhitt (approved):

PR #170, opened 2024-07-16
feat(ao): allow AO env variables to point process evaluation against …

dtfiedler:
…specific AO infrastrucutre

The defaults are reccomended, but this gives operators the ability to point the resolvers dryRuns to specific AO infrastructure. Useful for testing and if you are running any part of an AO stack yourself to avoid rate-limiting/network issues.

### DIFF
--- a/.github/workflows/build-core.yml
+++ b/.github/workflows/build-core.yml
@@ -27,7 +27,7 @@ jobs:
         if: matrix.os == 'macos-latest'
 
       # Build and test TypeScript
-      - run: yarn --immutable --immutable-cache
+      - run: yarn --immutable --immutable-cache --ignore-engines
       - run: yarn build
       - run: yarn lint:check
       - run: yarn test:ci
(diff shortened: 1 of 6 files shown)


.github/workflows/build-core.yml
```diff
@@ -27,7 +27,7 @@ jobs:
         if: matrix.os == 'macos-latest'
 
       # Build and test TypeScript
-      - run: yarn --immutable --immutable-cache
+      - run: yarn --immutable --immutable-cache --ignore-engines
```
dtfiedler, line 30: aoconnect requires this 😢 - they rec npm over yarn so installs fail

karlprieb (approved):

.github/workflows/build-core.yml
karlprieb: sad...

djwhitt (approved):

PR #173, opened 2024-07-17
feat(routes arweave): add minimal chunk post endpoint PE-5692

karlprieb:
Adds a minimal chunk POST endpoint that receives chunks and sends them
to the trusted node. As is, this provides limited utility, but it will
serve as a starting point for broadcasting to multiple Arweave nodes.
Also note, the Envoy routes have not been changed, so testing this route
requires connecting directly to the core service.


src/arweave/composite-client.ts
```diff
@@ -690,4 +701,74 @@ export class ArweaveCompositeClient
...
+
+          return {
+            success: true,
+            statusCode: resp.status,
+            canceled: false,
+            timedOut: false,
+          };
+        } catch (error: any) {
+          let canceled = false;
+          let timedOut = false;
```
karlprieb, line 744: @djwhitt maybe default to undefined?

src/types.d.ts
```diff
@@ -402,6 +402,15 @@ export interface JsonChunk {
...
 }
 
+// TODO better name
+export interface JsonChunkPost {
+  data_root: string;
+  chunk: string;
+  data_size: string;
+  data_path: string;
+  offset: string;
+}
```
karlprieb, line 412: @djwhitt I think the name is fine

src/types.d.ts
```diff
@@ -446,6 +455,34 @@ export interface ChunkDataByAnySource {
...
+  timedOut: boolean;
+};
+
+interface BroadcastChunkResult {
+  successCount: number;
+  failureCount: number;
+  results: BroadcastChunkResponses[];
+}
+
+// TODO better name?
```
karlprieb, line 471: @djwhitt same here

djwhitt (approved):

PR #179, opened 2024-07-26
docs(adrs): init WIP ClickHouse ADR

djwhitt:
This ADR is not finished, but is open for comment, especially wrt what about this decision needs the most explanation.


djwhitt:
@coderabbitai pause

PR #182, opened 2024-08-06
feat: add option to index data item signature

karlprieb:
Add signature fetcher to fetch signatures from data items.
It will look for data items signatures in the indexer and if not found, it will fetch them from data sources.


src/arweave/composite-client.ts
```diff
@@ -642,14 +643,25 @@ export class ArweaveCompositeClient
     };
   }
 
-  async getData({ id }: { id: string }): Promise<ContiguousData> {
+  async getData({
+    id,
+    region,
+  }: {
+    id: string;
+    region?: Region;
```
dtfiedler, line 651: what's the reasoning for using `region` instead of `range` here?

karlprieb: We're using `region` in other places like tx-chunk. I just kept the "convention"

src/data/tx-chunks-data-source.ts
```diff
@@ -123,6 +127,21 @@ export class TxChunksDataSource implements ContiguousDataSource {
...
       });
 
+      if (region) {
+        const byteRangeStream = new ByteRangeTransform(
+          region.offset,
+          region.size,
+        );
+        return {
+          stream: stream.pipe(byteRangeStream),
+          size: region.size,
```
karlprieb, line 137: @dtfiedler @djwhitt I'm not sure if I can do this. Returning just a small part will not break indexing?

src/arweave/composite-client.ts
djwhitt: I used `Region` instead of `Range` because, in my mind, a range should specify `start` and `end`, and we're specifying `offset` and `size`.

djwhitt: I was surprised to see this was needed too. My intent with the multi-range request ticket was that we implement this (passing regions down) there and use that as part of the multi-range handling. The multi-range handling we implemented does seem relatively efficient though, so I'm not too concerned atm.

migrations/2024.08.01T15.22.49.bundles.nullable-data-items-signature.sql
```diff
@@ -0,0 +1,83 @@
+CREATE TABLE stable_data_items_new (
```
djwhitt, line 1: Instead of recreating the tables, we need to see if the second answer (the one using `PRAGMA writable_schema`)  described in this StackOverflow question (https://stackoverflow.com/questions/4007014/alter-column-in-sqlite) will work.

src/data/signature-fetcher.ts
```diff
@@ -0,0 +1,90 @@
...
+
+      let signatureBuffer = Buffer.alloc(0);
+
+      for await (const chunk of stream) {
+        signatureBuffer = Buffer.concat([signatureBuffer, chunk]);
+      }
+
+      return toB64Url(signatureBuffer);
+    } catch (error) {
+      this.log.error('Error fetching data itemsignature', {
```
djwhitt, line 83: Typo here in the log message: "itemsignature"

djwhitt (changes requested):
Generally looks good, but we need to try a different approach for the migration if possible (see inline comment) and adjust the handling of `/tx/<id>/data` requests (see inline comment for this too).

src/routes/graphql/resolvers.ts
```diff
@@ -82,6 +83,18 @@ export function resolveTxBundledIn(tx: GqlTransaction) {
...
+export async function resolveTxSignature(tx: GqlTransaction) {
+  if (tx.signature !== null) {
+    return tx.signature;
+  }
+
+  if (tx.parentId !== null) {
+    return signatureFetcher.getDataItemSignature(tx.id);
+  }
+
+  return '';
```
karlprieb, line 95: return '<not-found>' instead

src/arweave/composite-client.ts
```diff
@@ -642,14 +643,25 @@ export class ArweaveCompositeClient
...
     this.failureSimulator.maybeFail();
 
     try {
       const [dataResponse, dataSizeResponse] = await Promise.all([
         this.trustedNodeRequestQueue.push({
           method: 'GET',
           url: `/tx/${id}/data`,
+          headers: region
+            ? {
+                Range: `bytes=${region.offset}-${region.offset + region.size - 1}`,
```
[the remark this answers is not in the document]
karlprieb: Fixed here [`a1172af` (#182)](https://github.com/ar-io/ar-io-node/pull/182/commits/a1172af5960978394927777ac15a50f40f4ac0d2)

src/routes/graphql/resolvers.ts
karlprieb: Fixed here [`05a68d4` (#182)](https://github.com/ar-io/ar-io-node/pull/182/commits/05a68d4c07097d5dde47e5333ed6609ed3ad83be)

src/data/signature-fetcher.ts
karlprieb: Fixed here [`05a68d4` (#182)](https://github.com/ar-io/ar-io-node/pull/182/commits/05a68d4c07097d5dde47e5333ed6609ed3ad83be)

migrations/2024.08.01T15.22.49.bundles.nullable-data-items-signature.sql
karlprieb: As we discussed, we will keep the migration this way as it's safest this way.

djwhitt (approved):

PR #187, opened 2024-08-20
feat: add option to index transaction signature

karlprieb:


src/database/standalone-sqlite.ts
```diff
@@ -629,6 +629,10 @@ export class StandaloneSqliteDatabaseWorker {
         }
 
         for (const tx of txs) {
+          console.log({ write: config.WRITE_TRANSACTION_DB_SIGNATURES });
```
djwhitt, line 632: Looks like some debug logging was left in here.

src/data/signature-fetcher.ts
```diff
@@ -84,6 +89,40 @@ export class SignatureFetcher implements SignatureSource {
...
+
+      return undefined;
+    }
+  }
+
+  async getTransactionSignature(id: string): Promise<string | undefined> {
+    try {
+      this.log.debug('Fetching transaction signature', { id });
+
+      const dataAttributes = await this.dataIndex.getDataAttributes(id);
```
djwhitt, line 101: I think we should make a new `getTxAttributes` method for this rather than adding the signature `getDataAttributes`. I both don't want to pay the cost of all the queries in `getDataAttributes` here and don't want to add the overhead of fetching signatures in `getDataAttributes`.

djwhitt (changes requested):

src/data/signature-fetcher.ts
karlprieb: Make sense
[`240c0c4` (#187)](https://github.com/ar-io/ar-io-node/pull/187/commits/240c0c47d769d31de170db688e5dff81f17fc1a7)

djwhitt: This adds the new method, but it looks like `signature` is still being returned from `getDataAttributes`. Unless we need that elsewhere, let's stop returning it from `getDataAttributes` (both the method and query results).

karlprieb: [`229ada5` (#187)](https://github.com/ar-io/ar-io-node/pull/187/commits/229ada578e067cf229c52c15e6c75c571389d6d2)

djwhitt (approved):

djwhitt (approved):

PR #190, opened 2024-08-27
feat(Ans104Unbundler): skip bundles when data item indexing queue is too big

karlprieb:


src/system.ts
```diff
@@ -361,13 +361,31 @@ export const contiguousDataSource = new ReadThroughDataCache({
...
+
+const ans104DataIndexer = new Ans104DataIndexer({
+  log,
+  eventEmitter,
+  indexWriter: nestedDataIndexWriter,
+  contiguousDataIndex,
+});
+
+const shouldUnbundleDataItems = () =>
+  ans104DataIndexer.queueDepth() < config.MAX_DATA_ITEM_QUEUE_SIZE ||
```
djwhitt, line 378: Let's make this `&&` so that we unbundle only when both are below the threshold.

src/workers/ans104-unbundler.ts
```diff
@@ -66,33 +69,46 @@ export class Ans104Unbundler {
...
   ): Promise<void> {
+    const log = this.log.child({ method: 'queueItem', id: item.id });
+
     if (this.workerCount === 0) {
+      log.warn('Skipping data item queuing due to no workers.');
       return;
     }
-    const log = this.log.child({ method: 'queueItem', id: item.id });
+
+    if (!this.shouldUnbundle()) {
```
karlprieb, line 107: Here I'm skipping it even if it's prioritized. Should we keep that way or ignore `shouldUnbundle` when the item is prioritized?

djwhitt: Good question. I'd say let's ignore it if it's prioritized `shouldUnbundle`.

src/system.ts
karlprieb: [`e82c796` (#190)](https://github.com/ar-io/ar-io-node/pull/190/commits/e82c796a875b57a83859a90fc47c8b6c3ddf10c3)

src/workers/ans104-unbundler.ts
karlprieb: [`dfe19de` (#190)](https://github.com/ar-io/ar-io-node/pull/190/commits/dfe19de7fa8e785a0ada4394bfad416c384a9151)

djwhitt (approved):

PR #202, opened 2024-09-12
fix(arns): return right away on arns HEAD requests

dtfiedler:
Avoid fetching data when handling HEAD requests for arns names

### DIFF
--- a/src/middleware/arns.ts
+++ b/src/middleware/arns.ts
@@ -19,7 +19,7 @@ import { Handler } from 'express';
 import { asyncMiddleware } from 'middleware-async';
 
 import * as config from '../config.js';
-import { headerNames } from '../constants.js';
+import { headerNames, REQUEST_METHOD_HEAD } from '../constants.js';
 import { sendNotFound } from '../routes/data/handlers.js';
 import { DATA_PATH_REGEX } from '../constants.js';
 import { NameResolution, NameResolver } from '../types.js';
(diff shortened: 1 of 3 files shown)


src/middleware/arns.ts
```diff
@@ -105,5 +105,11 @@ export const createArnsMiddleware = ({
     res.header(headerNames.arnsProcessId, processId);
     // TODO: add a header for arns cache status
     res.header('Cache-Control', `public, max-age=${ttl}`);
+
+    if (req.method === REQUEST_METHOD_HEAD) {
```
djwhitt, line 109: I'm unclear what we're fixing by doing this. We already terminate early and tear down the stream in the data handler if we get a `HEAD` request. Is that not working as intended? If return hear our responses won't include the correct `Content-Length` and `Content-Type`.

dtfiedler: AH, forgot about those headers. Yes, we still end up fetching the data item stream which are then used to set the headers (as implemented)- [here](https://github.com/ar-io/ar-io-node/blob/develop/src/routes/data/handlers.ts#L374-L395)

We do call `getDataAttributes` earlier in the handler. We could check the request method there and then set the data headers and return on HEAD requests before loading the stream. But you're right, we stop the stream right away on `HEAD` requests. So ultimately, this would avoid the fetching of the data item stream on HEAD requests.

dtfiedler: after some more discussion, we will close this PR and:
- create a ticket to pass the request type to `getData()` call to avoid creating `GET` method and then just terminating the stream on `HEAD` requests. 
- create a ticket to introduce resolver API endpoints to allow resolving names directly without worrying about potential data fetching/caching behavior in these middlewares

PR #205, opened 2024-09-13
feat(parquet-exporter): init parquet exporter

karlprieb:
Initial implementation of the parquet exporter.
Currently only supports data item tags.


src/workers/parquet-exporter.ts
```diff
@@ -0,0 +1,186 @@
...
+    outputDir,
+  }: {
+    startHeight: number;
+    endHeight: number;
+    maxFileRows: number;
+    outputDir: string;
+  }) {
+    console.log(outputDir);
+
+    const sqliteQuery = `
```
karlprieb, line 65: use just stable data items

src/routes/ar-io.ts
```diff
@@ -219,3 +221,35 @@ arIoRouter.post(
...
+
+      if (!outputDir || !startHeight || !endHeight || !maxFileRows) {
+        res.status(400).send('Missing required parameters');
+        return;
+      }
+
+      const parquetExporter = new ParquetExporter({
+        log,
+        duckDbPath: './data/duckdb/tags.duckdb',
+        sqliteDbPath: './data/sqlite/bundles.db',
```
dtfiedler, line 240: nit: put these behind env vars

src/workers/parquet-exporter.ts
```diff
@@ -0,0 +1,186 @@
...
+      './src/database/duckdb/schema.sql',
+      'utf8',
+    );
+
+    this.db.run(duckDbSchema);
+    this.log.debug('DuckDB schema imported successfully');
+
+    this.db.exec(`INSTALL sqlite; LOAD sqlite;`);
+    this.db.exec(`ATTACH '${sqliteDbPath}' AS sqlite_db (TYPE sqlite);`);
+    this.log.debug('SQLite database attached successfully');
```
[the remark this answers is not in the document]
dtfiedler: interesting - so with duckdb we load in sqlite, then can query on it using the duckdb driver? below we then query sqlite, inserting the results to duckdb to create the parquet export? very meta

src/routes/ar-io.ts
karlprieb: I didn't add a env var specifically for it as we handle paths a little bit different, but I added `DUCKDB_DATA_PATH` on docker-compose

src/database/duckdb/schema.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE IF NOT EXISTS tags (
+  height UBIGINT NOT NULL,
+  id BLOB NOT NULL,
+  tag_index USMALLINT NOT NULL,
+  created_at UBIGINT NOT NULL,
+  tag_name BLOB NOT NULL,
+  tag_value BLOB NOT NULL,
+  is_data_item BOOLEAN NOT NULL
```
karlprieb, line 8: @djwhitt 

I'm using UBIGINT for `height` and `created_at` so we don't need to worry about this numbers in the future.

For `tag_index` I'm using USMALLINT, but I think we will have more than 65k data items/bundle in a near future, right? Maybe we should change that to a UINT?

djwhitt: I believe `tag_index` is the index of the tag within the data item as opposed to within the bundle.

karlprieb: You're right 🤦

src/workers/parquet-exporter.ts
```diff
@@ -0,0 +1,197 @@
...
+      if (rowCount >= maxFileRows || height === maxHeight) {
+        const fileName = `${tableName}-minHeight:${minHeight}-maxHeight:${height}-rowCount:${rowCount}.parquet`;
+        const filePath = `${outputDir}/${fileName}`;
+
+        await this.db.exec(`
+        COPY (
+          SELECT * FROM ${tableName}
+          WHERE height >= ${minHeight} AND height <= ${height}
+        ) TO '${filePath}' (FORMAT PARQUET, COMPRESSION 'zstd')
+      `);
```
djwhitt, line 152: Trivial nit: indentation seems off here.

djwhitt (approved):
I believe I have found the smallest issue possible. :laughing: Looks good otherwise.

PR #207, opened 2024-09-18
cache for saveNestedDataHash calls to prevent duplicate writes

karlprieb:
Add a cache for saveNestedDataHash calls to prevent duplicate writes. Add a metrics counter for the number of duplicate saveNestedDataHash calls.

In an environment with ~830 calls to saveNestedDataHash per second after 7 minutes of operation(cache entry TTL), the amount of cached keys was ~349k, with a total of ~33MB of data stored in the cache.

Total amount of duplicated calls was 1428.

### DIFF
--- a/src/database/standalone-sqlite.ts
+++ b/src/database/standalone-sqlite.ts
@@ -29,6 +29,7 @@ import * as R from 'ramda';
 import sql from 'sql-bricks';
 import * as winston from 'winston';
 import CircuitBreaker from 'opossum';
+import NodeCache from 'node-cache';
 
 // TODO enable eslint
 /* eslint-disable */
@@ -2432,6 +2433,8 @@ export class StandaloneSqliteDatabase
     Awaited<ReturnType<StandaloneSqliteDatabase['getTransactionAttributes']>>
   >;
 
+  private saveNestedDataHashCache: NodeCache;
+
   constructor({
     log,
     coreDbPath,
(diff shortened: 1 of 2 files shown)


src/database/standalone-sqlite.ts
```diff
@@ -1151,6 +1154,19 @@ export class StandaloneSqliteDatabaseWorker {
         indexed_at: currentUnixTimestamp(),
       });
     }
+
+    if (this.insertDataHashCache.get(id)) {
```
djwhitt, line 1158: We should be using `hash` as the key here.

src/database/standalone-sqlite.ts
```diff
@@ -2924,6 +2968,17 @@ export class StandaloneSqliteDatabase
     parentId: string;
     dataOffset: number;
   }): Promise<void> {
+    const key = `${hash}:${parentId}`;
+
+    if (this.saveNestedDataHashCache.get(key)) {
```
djwhitt, line 2973: PiKNiK testing matches your test results, @karlprieb. We see some bursts of duplicates, but overall there aren't many here. Let's remove the deduplication here and save some memory.

src/database/standalone-sqlite.ts
```diff
@@ -1151,6 +1154,19 @@ export class StandaloneSqliteDatabaseWorker {
         indexed_at: currentUnixTimestamp(),
       });
     }
+
+    if (this.insertDataHashCache.get(hash)) {
```
djwhitt, line 1158: Did some manual logging of this one on PiKNiK and it looks effective.

djwhitt (changes requested):
2 out of 3 look effective. Let's remove the deduplication for `saveNestedDataHash`, but keep the rest.

djwhitt (approved):

PR #208, opened 2024-09-20
perf(arns): cache resolution for non-existent name

djwhitt:
Without this it's easy for non-existent names to DOS the CUs we're using to resolve names. This change caches non-existent name resolutions for 5 minutes in the in-memory/Redis cache.

### DIFF
--- a/src/resolution/on-demand-arns-resolver.ts
+++ b/src/resolution/on-demand-arns-resolver.ts
@@ -67,6 +67,7 @@ export class OnDemandArNSResolver implements NameResolver {
     });
     this.networkProcess = networkProcess;
     this.ao = ao;
+    // TODO: use getRecords instead of getArNSRecord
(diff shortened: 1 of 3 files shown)


src/resolution/on-demand-arns-resolver.ts
```diff
@@ -90,8 +91,18 @@ export class OnDemandArNSResolver implements NameResolver {
       // find that name in the network process, using the circuit breaker if there are persistent AO issues
       const arnsRecord = await this.aoCircuitBreaker.fire({ name: baseName });
 
-      if (arnsRecord === undefined || arnsRecord.processId === undefined) {
-        throw new Error('Invalid name, arns record not found');
+      if (
+        arnsRecord === undefined ||
```
arielmelendez, line 95: Keep throwing on `arnsRecord === undefined`

arielmelendez (approved):

PR #217, opened 2024-10-03
feat(parquet-exporter): add status endpoint for parquet export

karlprieb:
The new `/ar-io/admin/export-parquet/status` endpoint will return a json message with the following structure:
```
{
  pending: boolean;
  outputDir?: string;
  startHeight?: number;
  endHeight?: number;
  maxFileRows?: number;
  durationInSeconds?: number;
  error?: string;
}
```

`pending` is always present and the other attributes are present when needed.

### DIFF
--- a/src/workers/parquet-exporter.ts
+++ b/src/workers/parquet-exporter.ts
@@ -97,8 +97,9 @@ export class ParquetExporter {
     maxFileRows: number;
   }): Promise<void> {
     if (this.exportStatus.status === RUNNING) {
-      this.log.error('An export is already in progress');
-      return;
+      const error = new Error('An export is already in progress');
+      this.log.error(error.message);
+      return Promise.reject(error);
     }
 
     this.exportStatus.status = RUNNING;
--- a/src/workers/parquet-exporter.ts
+++ b/src/workers/parquet-exporter.ts
@@ -39,8 +39,15 @@ type Message = {
   stack?: string;
 };
 
-type ExportStatus = {
-  pending: boolean;
+type ExportStatus = 'not_started' | 'running' | 'completed' | 'errored';
+
+const NOT_STARTED: ExportStatus = 'not_started';
+const RUNNING: ExportStatus = 'running';
+const COMPLETED: ExportStatus = 'completed';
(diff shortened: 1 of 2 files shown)


src/workers/parquet-exporter.ts
```diff
@@ -114,36 +136,59 @@ export class ParquetExporter {
             durationInSeconds,
           });
 
-          this.isExporting = false;
-          this.worker?.terminate();
```
karlprieb, line 118: I stop terminating the worker directly from here because it runs before it closes the db and delete the files.

src/workers/parquet-exporter.ts
```diff
@@ -114,36 +136,59 @@ export class ParquetExporter {
...
+          this.exportStatus = {
+            pending: false,
+            outputDir,
+            startHeight,
+            endHeight,
+            maxFileRows,
+            durationInSeconds,
```
djwhitt, line 145: Let's include `endTime` too.

src/workers/parquet-exporter.ts
```diff
@@ -66,11 +94,12 @@ export class ParquetExporter {
...
     maxFileRows: number;
   }): Promise<void> {
-    if (this.isExporting) {
+    if (this.exportStatus.status === RUNNING) {
       this.log.error('An export is already in progress');
       return;
     }
-    this.isExporting = true;
+
+    this.exportStatus.status = RUNNING;
```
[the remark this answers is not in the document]
karlprieb: Good suggestion!

karlprieb:
@djwhitt I changed the `pending` property to `status` as we discussed in Slack [`5b5299e` (#217)](https://github.com/ar-io/ar-io-node/pull/217/commits/5b5299e88bac1f84508d27a8baffd779cb7f300f)

src/workers/parquet-exporter.ts
karlprieb: [`9237837` (#217)](https://github.com/ar-io/ar-io-node/pull/217/commits/9237837d114c3ebc949c9b26d3cee4e8e05bd025)

djwhitt (approved):

PR #218, opened 2024-10-04
feat: init seeding worker

hlolli:
- manage it with env-var based filters
 - add webtorrent dependency
 - implement method seed to SeedingWorker class


vilenarios:
Closing during a sweep of the open PR backlog. Two years open, currently conflicting against `develop`, and marked "experimenting" from the start.

The deciding factor is direction rather than age. Index distribution between gateways is being built as an **HTTP publish/subscribe index swarm** — see #914, which adds `src/index-swarm/` with a publisher, subscriber, gateway registry and CDB64 band kinds, distributing index data over HTTP rather than over a torrent swarm. That's the path being taken, so a webtorrent-based seeding worker isn't on it.

For the record, the parts of this that look like they'd still be worth something if seeding ever comes back: the env-var driven filter approach for deciding what to seed, and the `SeedingWorker` shape itself. The webtorrent dependency is the part that dates it.

Thanks @hlolli for the exploration — it was a reasonable thing to try at the time, and the question it was asking (how do gateways share large derived artifacts without everyone re-deriving them) is being answered, just differently.

PR #222, opened 2024-10-09
feat(data-attributes): save verified status

karlprieb:


src/database/standalone-sqlite.ts
```diff
@@ -2329,13 +2337,25 @@ export class StandaloneSqliteDatabaseWorker {
...
-      id: fromB64Url(id),
+      id: dataId,
       parent_id: fromB64Url(parentId),
       data_offset: dataOffset,
       data_size: dataSize,
-      indexed_at: currentUnixTimestamp(),
+      indexed_at: currentTimestamp,
     });
+
+    const parentAttributes = this.getDataAttributes(parentId);
```
karlprieb, line 2351: @djwhitt Is this the best way to check if parent is verified? If so I think we should create a cache for it as this query is expensive to run for each data item of the same parent

src/database/standalone-sqlite.ts
```diff
@@ -1140,26 +1140,34 @@ export class StandaloneSqliteDatabaseWorker {
...
     });
+
     if (dataRoot !== undefined) {
       this.stmts.data.insertDataRoot.run({
         data_root: fromB64Url(dataRoot),
         contiguous_data_hash: hashBuffer,
-        indexed_at: currentUnixTimestamp(),
+        indexed_at: currentTimestamp,
+        verified: verified ? 1 : 0,
+        verified_at: verified ? currentTimestamp : 0,
```
karlprieb, line 1170: null instead of 0

src/database/standalone-sqlite.ts
karlprieb: check on update query

src/database/standalone-sqlite.ts
```diff
@@ -2329,12 +2339,22 @@ export class StandaloneSqliteDatabaseWorker {
...
-      parent_id: fromB64Url(parentId),
+      id: idBuffer,
+      parent_id: parentIdBuffer,
       data_offset: dataOffset,
       data_size: dataSize,
-      indexed_at: currentUnixTimestamp(),
+      indexed_at: currentTimestamp,
+    });
+
+    this.stmts.data.updateVerifiedDataId.run({
```
karlprieb, line 2354: @djwhitt do you think it make sense to run this update in saveNestedDataId or just get the `verified` value from parent while running `insertDataId`?

djwhitt: Hmm... good question. I think it's probably better to do it in `insertDataId` and save a query.

karlprieb: [`6ccf079` (#222)](https://github.com/ar-io/ar-io-node/pull/222/commits/6ccf079d351126718a8023c7939b17a68974207d)

djwhitt (approved):

PR #225, opened 2024-10-21
feat: x-ar-io-root-transaction-id header

karlprieb:


src/database/sql/core/data-attributes.sql
```diff
@@ -1,19 +1,19 @@
...
   FROM stable_transactions
   WHERE id = @id
   UNION
-  SELECT data_root, data_size, content_type, content_encoding, false AS stable
+  SELECT data_root, data_size, content_type, content_encoding, null AS root_transaction_id, false AS stable
   FROM new_transactions
   WHERE id = @id
   UNION
-  SELECT null, data_size, content_type, content_encoding, true AS stable
+  SELECT null AS data_root, data_size, content_type, content_encoding, root_transaction_id, true AS stable
```
karlprieb, line 12: changed to `null AS data_root` for readability

src/database/standalone-sqlite.ts
```diff
@@ -981,12 +981,19 @@ export class StandaloneSqliteDatabaseWorker {
     const hash = dataRow?.hash;
     const dataRoot = coreRow?.data_root;
 
+    const rootTransactionId =
+      coreRow?.root_transaction_id !== null &&
+      coreRow?.root_transaction_id !== undefined
```
karlprieb, line 986: checking for `null` AND `undefined` because it can return null for transactions where `root_transaction_id` doesn't exist and return `undefined` for bundles/data items.

dtfiedler (approved):

djwhitt (approved):

PR #226, opened 2024-10-22
feat(fs): introduce chunk fs cache cleanup worker

dtfiedler:
Chunks can quickly fill up a disk, even if contiguous data and header filesystem caches are cleaned up. This PR introduces a chunk filesystem cleanup worker that is disabled by default and responsible for deleting old(er) chunks.


src/system.ts
```diff
@@ -224,6 +224,34 @@ export const contiguousDataFsCacheCleanupWorker = !isNaN(
...
+
+            return mostRecentTime <= thresholdDate;
+          } catch (err) {
+            log.error(`Error getting file stats for ${path}`, err);
+            return false;
+          }
+        },
+      })
+    : undefined;
+
```
thiagocarvalhodev, line 254: This block is almost identical to the `contiguousDataFsCacheCleanupWorker`. 

Consider abstracting the shared logic into a reusable function, for example:
```javascript
function createFsCacheCleanupWorker({ basePath, thresholdInSeconds, isEnabled = true }) {
  if (!isEnabled || isNaN(thresholdInSeconds)) {
    return undefined;
  }

  return new FsCleanupWorker({
    log,
    basePath,
    shouldDelete: async (path) => {
      try {
        const stats = await fs.promises.stat(path);
        const mostRecentTime =
          stats.atime > stats.mtime ? stats.atime : stats.mtime;
        const currentTimestamp = Date.now();
        const thresholdDate = new Date(
          currentTimestamp - thresholdInSeconds * 1000,
        );
        return mostRecentTime <= thresholdDate;
      } catch (err) {
        log.error(`Error getting file stats for ${path}`, err);
        return false;
      }
    },
  });
}

export const contiguousDataFsCacheCleanupWorker = createFsCacheCleanupWorker({
  basePath: 'data/contiguous',
  thresholdInSeconds: contiguousDataCacheCleanupThresholdInSeconds,
});

export const chunkDataFsCacheCleanupWorker = createFsCacheCleanupWorker({
  basePath: 'data/chunks',
  thresholdInSeconds: config.CHUNK_DATA_CACHE_CLEANUP_THRESHOLD,
  isEnabled: config.ENABLE_FS_CHUNK_CACHE_CLEANUP,
});
```

dtfiedler: indeed it is - good rec! will pull in

PR #228, opened 2024-10-31
feat: root_parent_offset and X-AR-IO-Data-Item-Data-Offset header

karlprieb:


test/end-to-end/data.test.ts
```diff
@@ -333,6 +333,69 @@ describe('X-AR-IO-Root-Transaction-Id header', function () {
...
+
+    // hopefully enough time to unbundle it
+    await wait(10000);
+  });
+
+  after(async function () {
+    await compose.down();
+  });
+
+  it('Verifying header for L1 bundle', async function () {
```
karlprieb, line 380: @djwhitt do you think I should query DB to check if the calculation was done properly?
I don't think it make sense as this bundle will never change. One thing that I could do is to describe the reasoning like:
bdi has `root_parent_offset` X and `data_offset` Y so data item data offset should be Z.

src/workers/ans104-unbundler.ts
```diff
@@ -37,6 +38,12 @@ interface IndexProperty {
 export type UnbundleableItem = (NormalizedDataItem | PartialJsonTransaction) &
   IndexProperty;
 
+const isNormalizedBundleDataItem = (
+  item: UnbundleableItem,
+): item is NormalizedBundleDataItem =>
+  (item as NormalizedBundleDataItem).root_parent_offset !== undefined &&
+  (item as NormalizedBundleDataItem).data_offset !== undefined;
```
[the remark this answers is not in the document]
karlprieb: Not possible because if root_parent_offset and data_offset are null it makes item to be NormalizedOptimisticDataItem and not NormalizedBundleDataItem

test/end-to-end/data.test.ts
djwhitt: @karlprieb Let's skip checking against the DB, but add a comparison of the data retrieved by requesting `Dc-q5iChuRWcsjVBFstEqmLTx4SWkGZxcVO9OTEGjkQ` by ID and using the size + offset. That seems like a good sanity check since that's the functionality we're ultimately trying to enable.

djwhitt (commented):
Functionality looks good. I added one suggestion about the test.

test/end-to-end/data.test.ts
karlprieb: @djwhitt here [`507bc89` (#228)](https://github.com/ar-io/ar-io-node/pull/228/commits/507bc892eb7d70c2edd6193976618b22c68daa8c)

djwhitt (approved):

PR #230, opened 2024-11-04
feat: data item indexing healthcheck

karlprieb:


djwhitt (commented):
Looks good. Testing on one our nodes now.

PR #233, opened 2024-11-12
feat(envoy): declare chunk post endpoint with retry policy

hlolli:


envoy/envoy.template.yaml
```diff
@@ -69,6 +69,32 @@ static_resources:
...
                             prefix_rewrite: '/'
+                        - match:
+                            prefix: '/chunk'
+                            headers:
+                              - name: ':method'
+                                exact_match: 'POST'
+                          route:
+                            cluster: ario_gateways
+                            timeout: 30s
+                            retry_policy: &post_retry_policy
```
djwhitt, line 80: We don't need a retry policy here. The ar-io-node service handles that.

envoy/envoy.template.yaml
```diff
@@ -69,6 +69,32 @@ static_resources:
...
+                              retriable_status_codes: 404
+                              retry_on: '5xx,reset,retriable-status-codes,connect-failure,refused-stream'
+                              num_retries: 3
+                              host_selection_retry_max_attempts: 3
+                              retry_host_predicate:
+                                name: envoy.retry_host_predicates.previous_hosts
+                                typed_config:
+                                  '@type': type.googleapis.com/envoy.extensions.retry.host.previous_hosts.v3.PreviousHostsPredicate
+                        - match:
+                            prefix: '/tx'
```
djwhitt, line 90: Why is this being added?

djwhitt (changes requested):

envoy/envoy.template.yaml
hlolli: removed

envoy/envoy.template.yaml
hlolli: removed

djwhitt (approved):

PR #238, opened 2024-11-20
feat(ArNSNamesCache): add a full arns names list cache

karlprieb:


src/resolution/composite-arns-resolver.ts
```diff
@@ -48,63 +50,74 @@ export class CompositeArNSResolver implements NameResolver {
     this.resolvers = resolvers;
     this.cache = cache;
     this.overrides = overrides;
+    this.arnsNamesCache = new ArNSNamesCache({ log });
```
karlprieb, line 53: I decided to initialize `ArNSNamesCache` here as we will only want to have it working if `CompositeArNSResolver` exists.

src/resolution/arns-names-cache.ts
```diff
@@ -0,0 +1,119 @@
...
+    networkProcess = IO.init({
+      process: new AOProcess({
+        processId: config.IO_PROCESS_ID,
+        ao: ao,
+      }),
+    }),
+    cacheTtl = DEFAULT_CACHE_TTL,
+  }: {
+    log: winston.Logger;
+    networkProcess?: AoIORead;
```
arielmelendez, line 56: Nit: type ordering differs from destructured ordering.

src/resolution/arns-names-cache.ts
```diff
@@ -0,0 +1,119 @@
...
+    const shouldRefresh = forceCacheUpdate || expiredTtl;
+
+    if (shouldRefresh) {
+      if (forceCacheUpdate) {
+        log.debug('Forcing cache update');
+      } else {
+        log.debug('Cache expired, refreshing...');
+      }
+      this.namesCache = this.getNamesFromContract();
+      this.lastCacheTime = Date.now();
```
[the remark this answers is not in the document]
karlprieb: It does make sense, but this will introduce an error if `getNames` is called before the first time cache is populated.

To workaround this we can init the class setting `lastCacheTime` to `Date.now()` but I'm not sure if it's a good approach.

src/resolution/composite-arns-resolver.ts
```diff
@@ -48,63 +50,74 @@ export class CompositeArNSResolver implements NameResolver {
...
-        ) {
-          metrics.arnsCacheHitCounter.inc();
-          this.log.info('Cache hit for arns name', { name });
-          return cachedResolution;
+    if (arnsNamesCache.length === 0) {
+      this.log.debug('Cached ArNS names list is empty');
+    }
+
+    if (arnsNamesCache.includes(name)) {
+      try {
```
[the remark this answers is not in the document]
karlprieb: Good suggestion!

src/resolution/arns-names-cache.ts
karlprieb: [`7af0154` (#238)](https://github.com/ar-io/ar-io-node/pull/238/commits/7af0154fc3b86b250a93879b641fdce8a09cc929)

src/config.ts
```diff
@@ -359,6 +359,11 @@ export const ARNS_ON_DEMAND_CIRCUIT_BREAKER_RESET_TIMEOUT_MS =
     `${5 * 60 * 1000}`, // 5 minutes
   );
 
+export const ARNS_NAMES_CACHE_TTL_SECONDS = +env.varOrDefault(
+  'ARNS_NAMES_CACHE_TTL_SECONDS',
+  `${60 * 60}`, // 1 hour
```
djwhitt, line 364: Let's default this to 5 minutes. We don't want it to be too long since that'll impact how long it takes new names to be available on gateways.

docker-compose.yaml
```diff
@@ -80,6 +80,7 @@ services:
       - ARNS_CACHE_TTL_SECONDS=${ARNS_CACHE_TTL_SECONDS:-3600}
       - ARNS_CACHE_MAX_KEYS=${ARNS_CACHE_MAX_KEYS:-10000}
       - ARNS_CACHE_TYPE=${ARNS_CACHE_TYPE:-redis}
+      - ARNS_NAMES_CACHE_TTL_SECONDS=${ARNS_NAMES_CACHE_TTL_SECONDS:-3600}
```
djwhitt, line 83: We should strip this default out of `docker-compose.yaml`. The default `config.ts` will handle a missing value. If you notice other obvious cases where we're doing that, feel free to correct them too, but we don't need to get them all in this PR.

src/config.ts
karlprieb: @dtfiedler and I discussed about this and we think 1 hour is a nice TTL as the goal is to reduce CU load with these requests and a name cannot be prescribed until the epoch after it has been purchased

dtfiedler: @karlprieb - i pull back my original suggestion of an hour. as name purchases ramp up this could cause some frustration of names not resolving quickly. 5 mins sounds good

src/resolution/arns-names-cache.ts
```diff
@@ -0,0 +1,119 @@
...
+  async getCacheSize(): Promise<number> {
+    const names = await this.namesCache;
+    return names.size;
+  }
+
+  private async getNamesFromContract(): Promise<Set<string>> {
+    const log = this.log.child({ method: 'getNamesFromContract' });
+    log.info('Starting to fetch names from contract');
+
+    const records = await fetchAllArNSRecords({
```
dtfiedler, line 104: this API is fine to use - we do the pagination manually in the ar-io-observer to have more control over failure scenarios. 

https://github.com/ar-io/ar-io-observer/blob/develop/src/names/contract-names-source.ts#L36-L50

docker-compose.yaml
karlprieb: [`02cfccf` (#238)](https://github.com/ar-io/ar-io-node/pull/238/commits/02cfccf790fb875be4a9837ac282d15583721d83)

src/resolution/arns-names-cache.ts
```diff
@@ -0,0 +1,119 @@
...
+      contract: this.networkProcess,
+    });
+
+    const names = new Set(Object.keys(records));
+
+    if (names.size === 0) {
+      // Assume empty results mean there was an error
+      // The current sdk version we use fetchAllArNSRecords does not throw an error
+      throw new Error('Failed to fetch ArNS records');
```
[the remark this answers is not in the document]
karlprieb: [`02cfccf` (#238)](https://github.com/ar-io/ar-io-node/pull/238/commits/02cfccf790fb875be4a9837ac282d15583721d83)

djwhitt (approved):

src/resolution/arns-names-cache.ts
djwhitt: @dtfiedler Based on this review, I'm planning to write up an ADR on ArNS resolution behavior. There are enough edge cases I think we need to write down what exactly we want and then bring the code in line with that. That said, what we have here seems like a reasonable next step given the top level fallback mechanisms we already have in place.

dtfiedler (approved):

fedellen (approved):

PR #247, opened 2024-12-10
feat: add import data from gql script

karlprieb:
This Nodejs script fetches ArDrive L1 transactions and bundles from a gateway through GQL interface.

The script can receive some parameters like:
  - --gqlEndpoint # defaults to Goldsky
  - --minHeight # defaults to 0
  - --maxHeight # defaults to the latest Arweave block
  - --blockRangeSize # defaults to 100

Example:
`node --import=./register.js scripts/import-data/fetch-data-gql.ts --minHeight 1000000 --maxHeight 1000500`


scripts/import-data/fetch-data-gql.ts
```diff
@@ -0,0 +1,329 @@
...
+const args = process.argv.slice(2);
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+
+let GQL_ENDPOINT = 'https://arweave-search.goldsky.com/graphql';
+let MIN_BLOCK_HEIGHT = 0;
+let MAX_BLOCK_HEIGHT: number | undefined;
+let BLOCK_RANGE_SIZE = 100;
+
+args.forEach((arg, index) => {
```
djwhitt, line 30: Definitely not required, but it would be fine to pull in Commander or similar for arg parsing.

djwhitt (commented):
We need an extra GQL call to get the root parent for BDIs. It also would be worthwhile to make the GQL query we're using to find data items configurable. 

I commented on a couple places where we could pull in libraries too, but perhaps the intent is to make this script as standalone as possible. I think that's worth doing for now, but if it becomes to cumbersome pulling in some libraries is fine too.

scripts/import-data/fetch-data-gql.ts
karlprieb: I wanted to keep things simple that's why I chose with `process.argv`

scripts/import-data/fetch-data-gql.ts
```diff
@@ -0,0 +1,329 @@
...
+        console.error('Missing value for --blockRangeSize');
+        process.exit(1);
+      }
+      break;
+    default:
+      break;
+  }
+});
+
+const fetchWithRetry = async (
```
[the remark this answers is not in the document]
karlprieb: I wanted to keep things simple and decided to not use external packages until I need something that is really complex.

scripts/import-data/fetch-data-gql.ts
djwhitt: Yeah, seems like a good idea. Then we can very easily copy/paste this script or point people at the file.

scripts/import-data/fetch-data-gql.ts
```diff
@@ -0,0 +1,329 @@
...
+
+      if (!transactions.has(blockHeight)) {
+        transactions.set(blockHeight, new Set());
+      }
+      if (!bundles.has(blockHeight)) {
+        bundles.set(blockHeight, new Set());
+      }
+
+      if (bundleId !== undefined) {
+        bundles.get(blockHeight)?.add(bundleId);
```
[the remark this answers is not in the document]
djwhitt: I was thinking it might be nice to skip the extra lookups and only get the BDIs for cases where someone wants the script to run faster.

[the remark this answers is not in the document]
karlprieb: So it would be a setting for bundles where you can get the root tx id OR the bdi id (current behavior in this PR), right?

scripts/import-data/count-fetched-ids.ts
```diff
@@ -0,0 +1,174 @@
...
+  let counter = 0;
+  for (const file of files) {
+    const filePath = path.join(folder, file);
+    const ids = JSON.parse(await fs.readFile(filePath, 'utf-8')) as string[];
+    counter += ids.length;
+  }
+  return counter;
+};
+
+// const importFromFiles = async ({
```
djwhitt, line 86: Looks like this could be removed.

djwhitt (approved):
Noticed some commented code that was left in that could be removed, but otherwise looks good.

scripts/import-data/count-fetched-ids.ts
karlprieb: Yep, my bad

djwhitt:
Marked my review as pending again since the Parquet export script will go into this branch too.

scripts/import-data/export-parquet.ts
```diff
@@ -0,0 +1,114 @@
...
+      startHeight: MIN_BLOCK_HEIGHT,
+      endHeight: MAX_BLOCK_HEIGHT,
+      maxFileRows: MAX_FILE_ROWS,
+    }),
+  });
+
+  console.log(
+    `Parquet export started from block ${MIN_BLOCK_HEIGHT} to ${MAX_BLOCK_HEIGHT}`,
+  );
+})();
```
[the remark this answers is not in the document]
karlprieb: IT will throw and error if POST fail

karlprieb:
@djwhitt We still can improve this set of scripts adding one to fetch the missing root tx ids. I'm not sure this is needed for this PR but it's something we could do.

djwhitt (approved):

PR #252, opened 2024-12-25
Replace on-failure with unless-stopped docker-compose.yaml

owlstake:
sometimes docker does not restart when it fails


dtfiedler:
Hi @owlstake - thanks for the issue. 

Before we review the change - can you provide more details on what container you're seeing not restart? If you have any logs of the container failing to restart as well that would be helpful. We may want to adjust the error codes we throw in the service rather than modify the docker restart behavior.

djwhitt:
Seems safe enough to do regardless, but I would still love to know what issue you're seeing that requires this, @owlstake.

djwhitt (approved):

PR #257, opened 2025-01-03
feat(post-chunk): post chunks to secondary urls

karlprieb:


src/arweave/composite-client.ts
```diff
@@ -726,67 +828,106 @@ export class ArweaveCompositeClient
...
+        secondaryResults.push(task);
+      }
+
+      results = [...results, ...(await Promise.all(secondaryResults))];
+    }
+
     return {
-      successCount,
+      successCount: successCount + secondarySuccessCount,
       failureCount,
```
karlprieb, line 931: @djwhitt should we add success and failure count from secondary urls here as well?

djwhitt: Good question. No, I think we should adjust the logic slightly to make sending to chunk nodes "fire and forget" rather than waiting so that we can return results.

src/arweave/composite-client.ts
```diff
@@ -726,67 +828,106 @@ export class ArweaveCompositeClient
...
+              canceled: error.code === 'ERR_CANCELED',
+              timedOut: error.code === 'ECONNABORTED',
+            };
+          }
+        });
+
+        secondaryResults.push(task);
+      }
+
+      results = [...results, ...(await Promise.all(secondaryResults))];
```
djwhitt, line 926: As mentioned below, let's adjust this to avoid waiting for results before returning a response.

djwhitt (approved):

PR #264, opened 2025-01-09
fix(arns): introduce cache miss/hit debouncing for arns name cache

dtfiedler:
Adds debouncing to the arns name list on cache hit/misses based on newly exposed variables. These should help alleviate name resolution delays for new names, as the cache will be refreshed within 10 seconds on misses (by default). Additionally, operators can choose to set cache hit debounce interval separate from the cache TTL, allowing tighter control of how often to refresh the cache - but always preserving it for at least the TTL if it fails to request an updated list from the CU. Refer to the updated ADR for specifics on how these caching mechanisms are architected.

Update:

Introducing a new `Kv-Debounce-Store` which accepts any existing `KvStore` implementation and a `hydrateFn` that is called after the provided debounce intervals.

### DIFF
--- a/src/resolution/arns-names-cache.ts
+++ b/src/resolution/arns-names-cache.ts
@@ -82,9 +82,9 @@ export class ArNSNamesCache {
       hydrateImmediately: true,
       /**
        * Bind the hydrateArNSNamesCache method to the ArNSNamesCache instance
-       * so that the hydrateFn has access to this instance's properties and methods (e.g. this.log, this.networkProcess, etc.).
+       * so that the debounceFn has access to this instance's properties and methods (e.g. this.log, this.networkProcess, etc.).
        */
-      hydrateFn: this.hydrateArNSNamesCache.bind(this),
+      debounceFn: this.hydrateArNSNamesCache.bind(this),
     });
   }
 
--- a/src/resolution/arns-names-cache.ts
+++ b/src/resolution/arns-names-cache.ts
@@ -22,37 +22,32 @@ import {
   AoARIORead,
   AOProcess,
   ARIO,
-  fetchAllArNSRecords,
-  AoArNSNameData,
+  AoArNSNameDataWithName,
+  PaginationResult,
 } from '@ar.io/sdk';
 import * as config from '../config.js';
 import { connect } from '@permaweb/aoconnect';
-
-const DEFAULT_CACHE_TTL = config.ARNS_NAMES_CACHE_TTL_SECONDS * 1000;
-const DEFAULT_MAX_RETRIES = 3;
-const DEFAULT_RETRY_DELAY = 5 * 1000; // 5 seconds
+import { KvDebounceStore } from '../store/kv-debounce-store.js';
+import { KVBufferStore } from '../types.js';
 
 const DEFAULT_CACHE_MISS_DEBOUNCE_TTL =
   config.ARNS_NAME_LIST_CACHE_MISS_REFRESH_INTERVAL_SECONDS * 1000;
 const DEFAULT_CACHE_HIT_DEBOUNCE_TTL =
   config.ARNS_NAME_LIST_CACHE_HIT_REFRESH_INTERVAL_SECONDS * 1000;
 
+/**
+ * Wraps an ArNS registry cache in a debounce cache that automatically refreshes
+ * the cache after the debounce ttl has expired.
+ */
 export class ArNSNamesCache {
   private log: winston.Logger;
   private networkProcess: AoARIORead;
-  private namesCache: Promise<Map<string, AoArNSNameData>>;
-  private lastSuccessfulNames: Map<string, AoArNSNameData> | null = null;
-  private lastCacheTime = 0;
-  private cacheTtl: number;
-  private maxRetries: number;
-  private retryDelay: number;
-  private cacheMissDebounceTtl: number;
-  private cacheHitDebounceTtl: number;
-  private debounceTimeout: NodeJS.Timeout | null = null;
-  private isDebouncing = false;
+  private arnsRegistryKvCache: KVBufferStore;
+  private arnsDebounceCache: KvDebounceStore;
 
   constructor({
     log,
+    registryCache,
     ao = connect({
       MU_URL: config.AO_MU_URL,
       CU_URL: config.AO_CU_URL,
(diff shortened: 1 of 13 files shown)


src/resolution/arns-names-cache.ts
```diff
@@ -89,20 +103,52 @@ export class ArNSNamesCache {
...
 
+  async has(name: string): Promise<boolean> {
+    const names = await this.getNames();
+    const nameExists = names.has(name);
+    // schedule the next debounce based on the cache hit or miss
+    await this.scheduleCacheRefresh(
+      nameExists ? this.cacheHitDebounceTtl : this.cacheMissDebounceTtl,
+    );
+    return nameExists;
+  }
```
dtfiedler, line 128: considering making this just `getName` instead of `has`

src/resolution/arns-names-cache.ts
```diff
@@ -89,20 +103,52 @@ export class ArNSNamesCache {
...
+
+  private async scheduleCacheRefresh(ttl: number): Promise<void> {
```
dtfiedler, line 130: this method could be added to an interface as we'll use it multiple times

src/resolution/arns-names-cache.ts
dtfiedler, line 1: Note: this is a WIP - i'll be replacing what's currently here with the implementation of `Kv-Debounce-Store`.

src/store/kv-debounce-store.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  async get(key: string): Promise<Buffer | undefined> {
+    const value = await this.kvBufferStore.get(key);
+    if (value === undefined) {
+      this.debounceHydrateOnMiss();
+      return undefined;
+    }
+    this.debounceHydrateOnHit();
+    return value;
+  }
```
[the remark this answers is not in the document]
dtfiedler: reasonable suggestion, but i'll let callers dictate and manage the logs they want to see by attaching them to the `hydrateFn` they provide to this class

src/store/kv-debounce-store.test.ts
```diff
@@ -0,0 +1,140 @@
...
+      assert.equal(callCount, 1); // not called again until the debounce ttl expires
+      assert.equal(result, undefined);
+
+      // // wait the cache miss debounce ttl and confirm hydrateFn was called again
+      await new Promise((resolve) => setTimeout(resolve, 10));
+      const result2 = await kvDebounceStore.get(key);
+      assert.equal(callCount, 2);
+      assert.ok(lastCallTimestamp >= Date.now() - 10);
+      assert.equal(result2!.toString('utf-8'), '1'); // eslint-disable-line @typescript-eslint/no-non-null-assertion
+    });
```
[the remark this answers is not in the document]
dtfiedler: [dab60a0](https://github.com/ar-io/ar-io-node/pull/264/commits/dab60a0853bb55b1a1363b6ad170fafc91394cc1)

src/store/kv-debounce-store.ts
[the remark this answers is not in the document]
hlolli: @coderabbai will the error object bubble up given that this.debounceHydrateOnMiss(); is a promise and the caller itself async get would also be returning a promise? The caller of await get(), would get the same error object as if calling this.debounceHydrateOnHit() promise directly, or would the "Error uncaught in promse" error show up?

src/store/kv-debounce-store.ts
```diff
@@ -0,0 +1,78 @@
...
+    // hydrate the cache immediately when the cache is created
+    if (hydrateImmediately) {
+      debounceFn();
+    }
+  }
+
+  async get(key: string): Promise<Buffer | undefined> {
+    const value = await this.kvBufferStore.get(key);
+    if (value === undefined) {
+      this.debounceHydrateOnMiss();
```
djwhitt, line 56: Did you consider awaiting the refresh on the miss?

djwhitt (commented):
Looking good overall. I like the refactoring. I added a few inline comments and questions. I also think we should add a Prometheus counter to track name list hits/misses independently from resolution hits/misses.

src/resolution/composite-arns-resolver.ts
```diff
@@ -74,61 +79,68 @@ export class CompositeArNSResolver implements NameResolver {
...
-            this.log.info('Cache hit for arns name', { name });
-            return cachedResolution;
-          }
+    try {
+      // check if our base name is in our arns names cache, this triggers a debounce with a ttl dependent on if it's in the cache or not
+      const baseNameInCache =
+        await this.arnsNamesCache.getCachedArNSBaseName(baseName);
+
+      if (!baseNameInCache) {
+        throw new Error('Base name not found in ArNS names cache');
```
dtfiedler, line 88: i will modify this to return the empty resolution data right away instead of throwing an error and a warning log

dtfiedler: [91cdc3d](https://github.com/ar-io/ar-io-node/pull/264/commits/91cdc3d186784a01c94f857b0ec0ec204369fa22)

src/store/kv-debounce-store.ts
dtfiedler: discussed on a call, concluded to await on the refresh. this means original request will wait the debounce period + the request time instead of having to request again. ultimately should help keep the cache fresh as requests come in and names expire from the internal KV cache

[91cdc3d](https://github.com/ar-io/ar-io-node/pull/264/commits/91cdc3d186784a01c94f857b0ec0ec204369fa22)

src/store/kv-arns-base-name-store.ts
```diff
@@ -0,0 +1,74 @@
...
+    this.kvBufferStore = kvBufferStore;
+    this.hashKeyPrefix = hashKeyPrefix;
+  }
+
+  private hashKey(key: string): string {
+    return `${this.hashKeyPrefix}|${key}`;
+  }
+
+  async get(key: string): Promise<Buffer | undefined> {
+    return this.kvBufferStore.get(this.hashKey(key));
```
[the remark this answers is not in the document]
dtfiedler: moved it to here - [91cdc3d](https://github.com/ar-io/ar-io-node/pull/264/commits/91cdc3d186784a01c94f857b0ec0ec204369fa22)

djwhitt (approved):
@dtfiedler Merging as is to get image build started. Metrics can be added/improved in the follow up.

PR #277, opened 2025-01-23
chore(docs): update docs/env.md file

dtfiedler:
This is mostly a GPT curated file generated from config.ts.


djwhitt:
@dtfiedler I'm holding off on merging this for now because I expect a bit more churn, but I'm curious what prompt(s) you used so I can replicate this alter when things settle down.

PR #284, opened 2025-01-24
WIP: Queue chunk downloads when data roots don't match

hlolli:


src/config.ts
```diff
@@ -239,9 +239,9 @@ export const MAX_DATA_ITEM_QUEUE_SIZE = +env.varOrDefault(
   '100000',
 );
 
-// The maximum number of bundles to queue for unbundling before skipping
-export const BUNDLE_DATA_IMPORTER_QUEUE_SIZE = +env.varOrDefault(
-  'BUNDLE_DATA_IMPORTER_QUEUE_SIZE',
+// The maximum number of data imports to queue for unbundling/importing before skipping
+export const DATA_IMPORTER_QUEUE_SIZE = +env.varOrDefault(
+  'DATA_IMPORTER_QUEUE_SIZE',
```
djwhitt, line 244: This data import will still be for bundles so the name shouldn't change. This may require a change to pass this value into the data importer if it's referenced directly in the data downloader.

monitoring/grafana/dashboards/examples/queues-example.json
```diff
@@ -856,13 +856,13 @@
...
           "editorMode": "builder",
-          "expr": "queue_length{job=\"ar_io_nodes\", queue_name=\"bundleDataImporter\"}",
+          "expr": "queue_length{job=\"ar_io_nodes\", queue_name=\"dataImporter\"}",
           "legendFormat": "{{instance}}",
           "range": true,
           "refId": "A"
         }
       ],
-      "title": "bundleDataImporter",
+      "title": "dataImporter",
```
djwhitt, line 865: This data import will still be for bundles so the name shouldn't change.

src/workers/data-verification.ts
```diff
@@ -63,6 +67,14 @@ export class DataVerificationWorker {
       workerCount,
       streamTimeout,
     });
+
+    this.dataImporter =
+      dataImporter ||
+      new DataImporter({
```
djwhitt, line 73: This should be instantiated in `system.ts` instead of here. It also needs to use a chunk based `contiguousDataSource` instead of the default background contiguous data source.

src/workers/data-importer.ts
```diff
@@ -76,47 +80,56 @@ export class BundleDataImporter {
...
+  }: DataImporterQueueItem): Promise<void> {
     const log = this.log.child({ method: 'download', id: item.id });
 
     const data = await this.contiguousDataSource.getData({ id: item.id });
 
     return new Promise((resolve, reject) => {
       data.stream.on('end', () => {
-        log.debug('Bundle data downloaded complete. Queuing for unbundling..');
-        this.ans104Unbundler.queueItem(item, prioritized, bypassFilter);
+        const isAnyContiguousData = this.isAnyContiguousData(item);
```
djwhitt, line 117: What is the purpose of `isAnyContiguousData`?

src/workers/data-importer.ts
```diff
@@ -76,47 +80,56 @@ export class BundleDataImporter {
...
     const log = this.log.child({ method: 'queueItem', id: item.id });
     if (this.workerCount === 0) {
-      log.debug('Skipping bundle download, no workers.');
+      log.debug('Skipping contiguous-data download, no workers.');
       return;
     }
 
     if (prioritized === true) {
-      log.debug('Queueing prioritized bundle download...');
+      log.debug('Queueing prioritized contiguous-data download...');
```
djwhitt, line 94: Nit: no '-' in contiguous data.

src/system.ts
```diff
@@ -472,15 +472,15 @@ metrics.registerQueueLengthGauge('ans104Unbundler', {
   length: () => ans104Unbundler.queueDepth(),
 });
 
-export const bundleDataImporter = new BundleDataImporter({
+export const dataImporter = new DataImporter({
```
djwhitt, line 475: The `bundelDataImporter` name should stay the same. Instead we should be instantiating a separate `DataImporter` that uses a chunk based contiguous data source and pass that to the verification worker.

djwhitt (commented):
Added some comments. Headed in the right direction, but we want to handle how we're instantiating and name data importers a bit differently.

src/workers/data-importer.ts
hlolli: in this case the item passed has just { id: string; } it's not the full data-item as we have in many places. I made a type for the two variants. From the verification side we don't have this data-item interface and just the tx-id. And the data-importer class cannot accept import data in that form as is. So it's a suggestion of how to solve it, and here isAnyContiguousData is just asserting that it's this { id: string; } format, which can be either layer-1 or layer-2 (maybe it means it's just layer-1 only because I think we only verify layer-1 data in the dataRootVerify, so a better variable name may be needed)

src/system.ts
hlolli: I see, that's a good idea, I will adjust. If having two sepearate classes would make sense too if the logic stats to diverge greatly enough. But I keep it 1 and same class as is called for, for now.

src/workers/data-importer.ts
hlolli: I'm reversing this logic, it should ask for isUnbundleableItem or not, that's an interface that's already present there.

src/workers/data-verification.ts
hlolli: done

src/system.ts
hlolli: done

src/workers/data-importer.ts
hlolli: fixed

src/workers/data-importer.ts
hlolli: done

monitoring/grafana/dashboards/examples/queues-example.json
hlolli: done (the dataImporter as prometheus metric exists separately but this file is now unchanged)

src/config.ts
hlolli: it's used in the class I think for the queue size of the worker jobs of the class, for both cases, but I did revert this still

src/system.ts
djwhitt: > If having two sepearate classes would make sense too if the logic stats to diverge greatly enough. But I keep it 1 and same class as is called for, for now.

Agree. I think we're good on that front for now, but if we get more conditionals in there we'll want to think about it more.

src/config.ts
djwhitt: > it's used in the class I think for the queue size of the worker jobs of the class, for both cases, but I did revert this still

Gotcha. I'll take a look when I rereview in a bit. We may want two constants now.

src/config.ts
```diff
@@ -239,7 +239,7 @@ export const MAX_DATA_ITEM_QUEUE_SIZE = +env.varOrDefault(
   '100000',
 );
 
-// The maximum number of bundles to queue for unbundling before skipping
+// The maximum number of data imports to queue for unbundling/importing before skipping
 export const BUNDLE_DATA_IMPORTER_QUEUE_SIZE = +env.varOrDefault(
```
djwhitt, line 243: Let's add another configuration option called `VERIFICATION_DATA_IMPORTER_QUEUE_SIZE` and use that for the verification download queue.

src/system.ts
```diff
@@ -472,7 +472,16 @@ metrics.registerQueueLengthGauge('ans104Unbundler', {
...
```
djwhitt, line 475: Minor tweak (sorry for the churn), but let's call this the `verificationDataImporter`.

src/workers/data-importer.ts
```diff
@@ -31,25 +31,27 @@ interface IndexProperty {
...
-export class BundleDataImporter {
+export class DataImporter {
   // Dependencies
   private log: winston.Logger;
   private contiguousDataSource: ContiguousDataSource;
-  private ans104Unbundler: Ans104Unbundler;
+  private ans104Unbundler: Ans104Unbundler | undefined;
 
-  // Unbundling queue
+  // contiguous-data queue
```
djwhitt, line 51: Nit: should be "Contiguous data"

src/workers/data-importer.ts
```diff
@@ -76,47 +80,56 @@ export class BundleDataImporter {
...
+      log.debug('Queueing prioritized contiguous data download...');
       this.queue.unshift({ item, prioritized, bypassFilter });
-      log.debug('Prioritized bundle download queued.');
+      log.debug('Prioritized contiguous data download queued.');
     } else if (this.queue.length() < this.maxQueueSize) {
-      log.debug('Queueing bundle download...');
+      log.debug('Queueing contiguous data download...');
       this.queue.push({ item, prioritized, bypassFilter });
-      log.debug('Bundle download queued.');
+      log.debug('contiguous data download queued.');
```
djwhitt, line 100: Nit: "contiguous" should be capitalized at the start of the sentence.

src/workers/data-verification.ts
```diff
@@ -129,6 +135,13 @@ export class DataVerificationWorker {
           indexedDataRoot,
           computedDataRoot,
         });
+        if (this.dataImporter) {
+          log.debug(
+            'Because of data-root mismatch, we are queueing data item for re-import.',
```
djwhitt, line 140: Nit: "re-import" should not contain a "-"

djwhitt (commented):
Added some tiny nits and I think we should add a new config option for the verification data queue size.

src/system.ts
hlolli: done, adjusted in the system.ts

src/workers/data-importer.ts
hlolli: fixed

src/workers/data-importer.ts
hlolli: fixed

src/workers/data-verification.ts
hlolli: fixed

src/config.ts
djwhitt: Don't worry about the docs. I'll go through and clean that up before release 24 (going to restructure `env.md` a bit too).

djwhitt (approved):

PR #285, opened 2025-01-24
chore(test): fix data verification test

dtfiedler:
We need to index the bundle transaction before we can attempting to verify its data root.

### DIFF
--- a/src/lib/data-root.ts
+++ b/src/lib/data-root.ts
@@ -160,7 +160,7 @@ export class DataRootComputer {
     });
   }
 
-  async computeDataRoot(id: string): Promise<Uint8Array | undefined> {
+  async computeDataRoot(id: string): Promise<string | undefined> {
     const tempPath = 'data/tmp/data-root';
     // eslint-disable-next-line no-async-promise-executor
     return new Promise(async (resolve, reject) => {
--- a/test/end-to-end/data-sources.test.ts
+++ b/test/end-to-end/data-sources.test.ts
@@ -65,7 +65,6 @@ describe('DataSources', () => {
 
       localStack = await new LocalstackContainer('localstack/localstack:3')
         .withNetwork(network as any)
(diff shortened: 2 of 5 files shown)


src/lib/data-root.ts
```diff
@@ -160,7 +160,7 @@ export class DataRootComputer {
     });
   }
 
-  async computeDataRoot(id: string): Promise<Uint8Array | undefined> {
+  async computeDataRoot(id: string): Promise<string | undefined> {
```
dtfiedler, line 163: @hlolli - the result of this is the `toB64Url` of the data root, lmk if i'm mistaken on this change. it was causing type warnings in the `verifyDataRoot` when trying to compare a `string` to `Uint8Array`

test/end-to-end/data-sources.test.ts
```diff
@@ -65,7 +65,6 @@ describe('DataSources', () => {
 
       localStack = await new LocalstackContainer('localstack/localstack:3')
         .withNetwork(network as any)
-        .withName('localstack')
```
dtfiedler, line 68: this was causing warnings/errors when attempting to spin up local stack. this will give it a dynamic name for each execution avoiding conflicts

```bash
Error: (HTTP code 409) unexpected - Conflict. The container name "/localstack" is already in use by container "420421cdc2b4abb71994d9034928ab0fd352c77cc1e0cc8fe05eff9112297135". You have to remove (or rename) that container to be able to reuse that name. 
```

src/lib/data-root.ts
hlolli: I think it was converted later, looking...

djwhitt (approved):

src/lib/data-root.ts
hlolli: that is correct, initially it was that, but after my changes it's a string

PR #286, opened 2025-01-24
feat(api): allow providing `bypassFilter` when manually enqueing a bu…

dtfiedler:
…ndle


src/routes/ar-io.ts
```diff
@@ -203,13 +203,18 @@ arIoRouter.post(
   express.json(),
   async (req, res) => {
     try {
-      const { id } = req.body;
+      const { id, bypassFilter = true } = req.body;
```
dtfiedler, line 206: implemented to original request, but we could also make `isPrioritized` available as a body param as well to validate that behavior

djwhitt: I think we can leave off `isPrioritized` for now. I am pondering another change though related to how we're queuing these.

test/end-to-end/indexing.test.ts
```diff
@@ -764,6 +767,72 @@ describe('Indexing', function () {
         assert.equal(dataItem.size, idAndSignatureType[id].size);
       });
     });
+
+    it('Verifying if request is rejected if byPassFilter is not a boolean', async function () {
+      const res = await axios.post(
+        'http://localhost:4000/ar-io/admin/queue-bundle',
+        {
+          id: 'FcWiW5v28eBf5s9XAKTRiqD7dq9xX_lS5N6Xb2Y89NY',
+          bypassFilter: 'true',
```
karlprieb, line 776: Shouldn't it be a boolean?

karlprieb (approved):

src/routes/ar-io.ts
```diff
@@ -218,7 +223,7 @@ arIoRouter.post(
       const queuedBundle = await system.queueBundle(
         { id, root_tx_id: id } as NormalizedDataItem | PartialJsonTransaction,
         true,
-        true,
+        bypassFilter,
```
djwhitt, line 226: @dtfiedler I realized after writing this ticket, that passing `bypassFilter: false` will be insufficient to cause filters to apply in the expected way since we have no metadata (tags, owner, etc.) associated with the ID when its queued here. I think what we should do instead is assume this is a base layer TX and queue the TX to be fetched by the `txFetcher` instead. Arguably, this is duplicate functionality (`/ar-io/admin/queue-tx` can accomplish this too), but I think it's probably still worth having the option given that `/ar-io/admin/queue-bundle` is most likely what people will reach for when they want to queue bundles.

djwhitt: cc @karlprieb for a sanity check on this.

djwhitt: @dtfiedler Still debating this a bit, so I'd say hold off any modifications for now. We could handle this via documentation too.

djwhitt: After some Slack discussion, I think we'll go for it. Queue to `txFetcher` instead when `bypassFilter` is set in the request.

dtfiedler: updated here - [7eb4dba](https://github.com/ar-io/ar-io-node/pull/286/commits/7eb4dbaccf9ef325621b064a3da04751d602d38c)

i put a note that we could alternatively redirect

djwhitt (approved):

PR #288, opened 2025-01-27
feat: weighted gateway peers for data retrievals

hlolli:
* adding 'peer' label to prometheus data retrival metrics
  * add lib helper 'random-weighted-choices' for returning weighted random list
  * add retry mechanism, retrying over random and unique gateway peers

### DIFF
--- a/src/data/ar-io-data-source.ts
+++ b/src/data/ar-io-data-source.ts
@@ -17,6 +17,10 @@
  */
 import { default as axios, AxiosResponse } from 'axios';
 import winston from 'winston';
+import {
+  WeightedElement,
+  randomWeightedChoices,
+} from '../lib/random-weighted-choices.js';
 import { AoARIORead } from '@ar.io/sdk';
 import { randomInt } from 'node:crypto';
 
@@ -31,6 +35,7 @@ import {
   parseRequestAttributesHeaders,
 } from '../lib/request-attributes.js';
 import * as metrics from '../metrics.js';
+import * as config from '../config.js';
 
 const DEFAULT_REQUEST_TIMEOUT_MS = 10_000; // 10 seconds
 const DEFAULT_UPDATE_PEERS_REFRESH_INTERVAL_MS = 3_600_000; // 1 hour
@@ -46,6 +51,8 @@ export class ArIODataSource implements ContiguousDataSource {
   private arIO: AoARIORead;
   peers: Record<string, string> = {};
   private intervalId?: NodeJS.Timeout;
+
+  protected weightedPeers: WeightedElement<string>[] = [];
 
   constructor({
     log,
@@ -121,6 +128,15 @@ export class ArIODataSource implements ContiguousDataSource {
       peers: Object.keys(peers),
     });
     this.peers = peers;
+    this.weightedPeers = Object.values(peers).map((id) => {
+      const previousWeight =
+        this.weightedPeers.find((peer) => peer.id === id)?.weight ?? undefined;
+      return {
+        id,
+        // the weight system is a bit arbitrary being between 0 and 100, 50 is the default neutral
+        weight: previousWeight === undefined ? 50 : previousWeight,
+      };
+    });
   }
 
   selectPeer(): string {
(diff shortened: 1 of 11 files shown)


src/data/ar-io-data-source.ts
```diff
@@ -185,86 +242,67 @@ export class ArIODataSource implements ContiguousDataSource {
...
-        return this.parseResponse(response, parsedRequestAttributes);
+        return this.parseResponse({
+          response,
+          requestAttributes: parsedRequestAttributes,
+          peer: currentPeer,
+        });
       } catch (error: any) {
         metrics.getDataErrorsTotal.inc({
-          class: this.constructor.name,
+          class: 'ArIODataSource',
```
djwhitt, line 276: Why switch back to hard coding here?

src/data/ar-io-data-source.ts
```diff
@@ -136,6 +151,43 @@ export class ArIODataSource implements ContiguousDataSource {
...
+
+  handlePeerSuccess(peer: string): void {
+    metrics.getDataStreamSuccessesTotal.inc({
+      class: this.constructor.name,
+      peer,
+    });
+    // warm the succeeding peer
+    this.weightedPeers.forEach((weightedPeer) => {
+      if (weightedPeer.id === peer) {
+        weightedPeer.weight = Math.min(weightedPeer.weight + 0.1, 100);
```
djwhitt, line 173: Let's make the warming and cooling values configurable.

src/lib/random-weighted-choices.ts
```diff
@@ -0,0 +1,98 @@
+/**
```
djwhitt, line 1: This algorithm seems good, but I'd suggest putting it through o1 and asking it to add detailed comments explaining the logic. We should also include a link to the original library and preserve its license here.

src/metrics.ts
```diff
@@ -263,19 +263,19 @@ export const arnsResolutionTime = new promClient.Summary({
 export const getDataErrorsTotal = new promClient.Counter({
   name: 'get_data_errors_total',
   help: 'Count of requests errors',
-  labelNames: ['class'],
+  labelNames: ['class', 'peer'],
```
djwhitt, line 266: We should confirm that the cases where this is incremented without specifying a `peer` still work.

djwhitt (commented):
Added a few comments, but overall this looks good.

I am curious if you have tested this at all to see how it performs? If you haven't, give it a try and see what happens (set `ON_DEMAND_RETRIEVAL_ORDER=ario-peer`).

Also, let's change the name of that data retrieval option to `ario-peers` but keep `ario-peer` working too for backwards compatibility.

src/data/ar-io-data-source.ts
hlolli: that's a mistake, I'll fix

src/lib/random-weighted-choices.ts
hlolli: I'll add a link to the package license field, it's the only documented license of the library, there's no license file or preamble present in the repo. Do we want to copy paste MIT license into the file, or just leave a link to the license field?
https://github.com/parmentf/random-weighted-choice/blob/626398ce68df3269996e05cc0d7e525f4a83322a/package.json#L40

src/metrics.ts
hlolli: I think peer will fail, I tried to enforce this via typescript by using as const

https://github.com/siimon/prom-client?tab=readme-ov-file#strongly-typed-labels

it does enforce that the keys are correctly named but they are still set as optional.

This metric get_data_errors_total and the other two are used in many places, often without peers (like just cache) or in places where the peer is unknown (because the subprocess only returns the response without the destination of the http request).

You would be fair to say that adjusting all callers to add peer is beyond the scope of this PR and we should look at alternative solutions. Which may include new metric or perhaps add a default to all the other callers, it is supposedly possible to set default label values with prom-client, (somthing we could look into) or we keep the diff Im about to push.

src/data/ar-io-data-source.ts
```diff
@@ -185,86 +242,67 @@ export class ArIODataSource implements ContiguousDataSource {
       }
     }
 
-    let selectedPeer = this.selectPeer();
+    const randomPeers = this.selectRandomWeightedPeers(totalRetryCount);
```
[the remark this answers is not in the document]
hlolli: done

src/data/ar-io-data-source.ts
hlolli: fixed

src/data/ar-io-data-source.ts
hlolli: I introduced env-var WEIGHTED_PEERS_TEMPERATURE_DELTA

src/lib/random-weighted-choices.ts
hlolli: I had o1 add comments

djwhitt (commented):
Almost there, just added a few minor comments. I'll also give this one more thorough review when those are done and my brain is a bit fresher.

src/lib/random-weighted-choices.ts
hlolli: done

src/data/tx-chunks-data-source.ts
```diff
@@ -118,12 +118,16 @@ export class TxChunksDataSource implements ContiguousDataSource {
       });
 
       stream.on('error', () => {
-        metrics.getDataStreamErrorsTotal.inc({ class: this.constructor.name });
+        metrics.getDataStreamErrorsTotal.inc({
+          class: this.constructor.name,
+          peer: 'unknown',
```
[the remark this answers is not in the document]
hlolli: it's labeled as chunks as opposed to unknown now

PR #290, opened 2025-01-28
feat: Search Arweave network for chunks

hlolli:
* add method 'peerGetChunk'
  * attempt to fetch chunk from a random peer before trustedPeers
  * new prometheus metrics 'request_chunk_total' and 'get_chunk_total'
  * add env-var 'ARWEAVE_NODE_IGNORE_URLS' to remove specific peers from
    the arweave node peer list

### DIFF
--- a/src/arweave/composite-client.ts
+++ b/src/arweave/composite-client.ts
@@ -30,6 +30,10 @@ import CircuitBreaker from 'opossum';
 import { FailureSimulator } from '../lib/chaos.js';
 import { fromB64Url } from '../lib/encoding.js';
 import {
+  WeightedElement,
+  randomWeightedChoices,
+} from '../lib/random-weighted-choices.js';
+import {
   sanityCheckBlock,
   sanityCheckChunk,
   sanityCheckTx,
@@ -112,6 +116,7 @@ export class ArweaveCompositeClient
   // Peers
   private peers: Record<string, Peer> = {};
   private preferredPeers: Set<Peer> = new Set();
+  private weightedChunkPeers: WeightedElement<string>[] = [];
 
   // Block and TX promise caches used for prefetching
   private blockByHeightPromiseCache = new NodeCache({
@@ -256,6 +261,9 @@ export class ArweaveCompositeClient
   }
 
   async refreshPeers(): Promise<void> {
+    const log = this.log.child({ method: 'refreshPeers' });
+    log.debug('Refreshing peers...');
+
     try {
       const response = await this.trustedNodeRequest({
         method: 'GET',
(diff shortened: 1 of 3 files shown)


src/metrics.ts
```diff
@@ -278,6 +278,22 @@ export const getDataStreamSuccessesTotal = new promClient.Counter({
...
 });
 
+//
+// Chunk source metrics
+//
+
+export const requestChunkFromPeerTotal = new promClient.Counter({
+  name: 'request_chunk_total',
+  help: 'Count of each individual chunk http request, status can be error or success',
+  labelNames: ['status', 'class', 'method', 'peer'] as const,
```
djwhitt, line 288: Let's remove `FromPeer` from the variable name, change the `peer` label to `source` (similar to the get data metrics), add a `type` that's either `trusted` or `peer`, and use this to count requests to the trusted node as well.

djwhitt (commented):
Added some minor feedback but overall looks good. I'll do another more detailed review tomorrow with a fresher brain.

src/arweave/composite-client.ts
```diff
@@ -264,33 +271,59 @@ export class ArweaveCompositeClient
...
+              if (response.data.blocks / response.data.height > 0.9) {
+                this.preferredPeers.add(this.peers[peerHost]);
+              }
+            } catch (error) {
+              metrics.arweavePeerInfoErrorCounter.inc();
             }
-          } catch (error) {
-            metrics.arweavePeerInfoErrorCounter.inc();
+          } else {
+            this.log.info('Ignoring peer:', { peerHost });
```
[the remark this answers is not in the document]
hlolli: fixed

src/metrics.ts
hlolli: pushed

I'm counting this metric on getChunkByAny method for trusted node, using the url stored in this.trustedNodeUrl as source.

src/arweave/composite-client.ts
```diff
@@ -256,6 +261,8 @@ export class ArweaveCompositeClient
   }
 
   async refreshPeers(): Promise<void> {
+    this.log.debug('ArweaveCompositeClient.refreshPeers refresh called');
```
djwhitt, line 264: Fine to leave this, but let's create a child logger with the method name and change the message to "Refreshing peers...".

djwhitt (commented):
One tiny log improvement needed then this one is good.

src/arweave/composite-client.ts
hlolli: done

djwhitt (approved):

PR #291, opened 2025-01-29
fix: shutdown problems with db.stop when exit signal is received

hlolli:
* remove all queue.drained calls in stop methods
  * better error handling around pool/worker stops in db.stop


karlprieb (commented):
We should await drained for the workers that use fastq

> queue.drained() => Promise
> Wait for the queue to be drained. The returned Promise will be resolved when all tasks in the queue have been processed by a worker.

karlprieb (approved):

djwhitt (approved):

PR #297, opened 2025-01-30
chore(dependencies): update major versions of lmdb,prom-client & dev-tooling

hlolli:


djwhitt:
Planning to do most of these, but not until after release 24.

djwhitt:
> Planning to do most of these, but not until after release 24.

Correction. I did all the needed security updates and will wait on the rest till after TGE.

vilenarios:
Closing during a sweep of the open PR backlog. Open since January 2025, conflicting against `develop`, and +2,188/-3,149 across 16 files with most of that in `yarn.lock`.

Dependency maintenance has been happening incrementally instead. Dependabot is active on this repo — #902, #903 and #904 are open right now — and the packages this PR targeted have moved on independently since it was written, so its pinned versions are no longer the versions anyone would want.

A conflicted lockfile diff of this size from eight months ago costs more to rebase and re-verify than to redo, and redoing it as a series of focused bumps is both easier to review and easier to bisect if something regresses. That's the approach already in flight.

Thanks @hlolli — the work wasn't wasted, it just got overtaken by the incremental path.

PR #299, opened 2025-01-30
feat(bundles): retry based on retry attempts

karlprieb:


src/workers/bundle-repair-worker.ts
```diff
@@ -107,6 +108,7 @@ export class BundleRepairWorker {
       );
       for (const bundleId of bundleIds) {
         this.log.info('Retrying failed bundle', { bundleId });
+        await this.bundleIndex.saveBundleRetries(bundleId);
```
karlprieb, line 111: @djwhitt I'm awaiting here to make sure we update retry stats

djwhitt (approved):

PR #300, opened 2025-01-30
chore(dependencies): migrate apollo-server 3->4

hlolli:


vilenarios:
Closing during a sweep of the open PR backlog — but the underlying work is being picked up, not dropped. Explaining both parts.

**Why this can't be merged as-is.** It migrates to Apollo Server 4, and [Apollo Server 4 reached end-of-life on 26 January 2026](https://www.apollographql.com/docs/apollo-server/previous-versions). The version this repo still runs, `apollo-server-express@3`, [went end-of-life on 22 October 2024](https://www.apollographql.com/blog/apollo-server-3-end-of-life-extension). Landing this would move the GraphQL request path from one unsupported major to another. Apollo Server 5 is the only supported line, and it works with Express 4 through [`@as-integrations/express4`](https://www.npmjs.com/package/@as-integrations/express4), which provides the same `expressMiddleware` that AS4 had built in.

**A caveat on the mergeable status.** GitHub reports this branch `MERGEABLE` / `CLEAN`, which is misleading — it means the changed *regions* don't overlap, not that the result would be correct. The `package.json` here also carries `@ar.io/sdk@^3.4.0-alpha.3` (develop is on `^4.0.0`), `lmdb@^3.2.2` (develop is on `^2.9.4`), `@types/express@^5.0.0` (develop is on `^4.17.21`), and an `lru-cache: 8.x` resolution (develop uses `^11.2.2`). Merging cleanly and building correctly are different things here.

**What's being done instead.** A fresh migration to Apollo Server 5 against current `develop`. This PR stays useful as a reference for the shape of it — particularly `makeApolloServerMiddleware` and the `src/app.ts` wiring, which are close to what AS5 needs.

The scope has grown since January 2025 in one way worth recording for whoever reviews the replacement. `src/routes/graphql/resolver-signal.ts` didn't exist when this PR was written, and it is built on an Apollo Server 3 implementation detail — its comments cite `apollo-server-core/dist/runHttpQuery.js:166` and it uses a nested `state` holder specifically to survive Apollo's shallow-clone of the context object. AS4+ changed the context lifecycle, so that workaround has to be re-derived rather than ported. It governs request-cancellation detection, and `graphql_requests_total` is the denominator for the disconnect rate, so an error there would corrupt metrics quietly rather than fail loudly.

Thanks @hlolli — this was right at the time, and it's the thing that surfaced that we're on an EOL GraphQL server.

PR #302, opened 2025-01-31
fix: data verify not falling back from chunk reimport

hlolli:


djwhitt (approved):

src/workers/data-verification.ts
```diff
@@ -128,9 +128,17 @@ export class DataVerificationWorker {
       }
 
       const indexedDataRoot = dataAttributes.dataRoot;
-      const computedDataRoot = await this.dataRootComputer.computeDataRoot(id);
+      let computedDataRoot: string | undefined = undefined;
+      try {
+        computedDataRoot = await this.dataRootComputer.computeDataRoot(id);
+      } catch (error) {
+        log.debug('Error computing data root.', { error });
+      }
```
dtfiedler, line 136: nit: could do it in a one-liner
```suggestion
      const computedDataRoot: string | undefined = await this.dataRootComputer.computeDataRoot(id).catch((e) =>       log.debug('Error computing data root.', { error }));
```

hlolli: fantastic, I take it

```
      const computedDataRoot: string | undefined = await this.dataRootComputer
        .computeDataRoot(id)
        .catch((error) => {
          log.debug('Error computing data root.', { error });
          return undefined;
        });
   ```

hlolli: done, and I re-tested the code, it's working the same way still

dtfiedler (approved):

PR #311, opened 2025-02-04
wip: ECDSA signature support

hlolli:


djwhitt (commented):
Looks great. Just one suggestion re a comment we should add.

djwhitt (approved):

PR #317, opened 2025-02-12
feat: warm weighted peers based on network performance

hlolli:
* add new axios-timings-adapter
  * add new env-var GATEWAY_PEERS_REQUEST_WINDOW_COUNT
  * conditionally warm peer more based on ttfd and download rate
  * workaround npm resolutions issue by pinning lru-cache


karlprieb (changes requested):
This PR introduces a change in the way we get data through Axios. Now with `AxiosResponseWithTimings` we will always get a full data response as buffer. Before we were handling a stream.

This change will increase memory usage and will probably generate OOM errors for large data.

Knowing that we should keep handling streams and not `Buffer.concat` it.

PR #323, opened 2025-02-12
feat: reset bundle timestamp when filters change

karlprieb:


djwhitt:
Is there an end-to-end test we can update for this one?

djwhitt (approved):
Spent a bit of time evaluating query performance, but I think it's fine (or not worse) since the one side of the `OR` is essentially a constant. Approving but waiting to merge till after release 27.

PR #325, opened 2025-02-13
feat(filters): log the matching status of each filter

hlolli:
With debug logs enabled, we can debug complicated bundle indexing filters and see the tree of user provided filters and if they are matching or not. This can simplify debugging process for users with multiple filters or complicated match criteria.

EDIT: Im adding default argument to a logger that isn't passed to make it simpler to make tests and experiment in the cli. But if we pass the logger right then we get the trees correctly.

### DIFF
--- a/src/filters.ts
+++ b/src/filters.ts
@@ -17,15 +17,55 @@
  */
 import { b64UrlToUtf8, fromB64Url, sha256B64Url } from './lib/encoding.js';
 import { ItemFilter, MatchableItem } from './types.js';
+import { Logger } from 'winston';
+import defaultLogger from './log.js';
+
+const logMatchResult = ({
+  log,
+  item,
+  isMatching,
+}: {
+  log: Logger;
+  item: MatchableItem;
+  isMatching: boolean;
+}) => {
+  if (isMatching) {
+    log.silly('Filter matched', {
+      id: item.id,
+      height: item.height,
+      parent: item.parent_id,
+    });
+  } else {
+    log.silly('Filter did not match', {
+      id: item.id,
+      height: item.height,
+      parent: item.parent_id,
+    });
+  }
+};
 
 export class AlwaysMatch implements ItemFilter {
-  async match(_: MatchableItem): Promise<boolean> {
+  private log: Logger;
+
+  constructor(log: Logger = defaultLogger) {
+    this.log = log.child({ class: this.constructor.name });
+  }
+
+  async match(item: MatchableItem): Promise<boolean> {
+    logMatchResult({ log: this.log, item, isMatching: true });
     return true;
   }
 }
 
 export class NeverMatch implements ItemFilter {
-  async match(_: MatchableItem): Promise<boolean> {
+  private log: Logger;
+
+  constructor(log: Logger = defaultLogger) {
+    this.log = log.child({ class: this.constructor.name });
+  }
+
+  async match(item: MatchableItem): Promise<boolean> {
+    logMatchResult({ log: this.log, item, isMatching: false });
     return false;
   }
 }
@@ -33,20 +73,27 @@ export class NeverMatch implements ItemFilter {
 export class NegateMatch implements ItemFilter {
   private readonly filter: ItemFilter;
 
-  constructor(filter: ItemFilter) {
+  private log: Logger;
+
+  constructor(filter: ItemFilter, log: Logger = defaultLogger) {
     this.filter = filter;
+    this.log = log.child({ class: this.constructor.name });
   }
 
   async match(item: MatchableItem): Promise<boolean> {
-    return !(await this.filter.match(item));
+    const isMatching = !(await this.filter.match(item));
+    logMatchResult({ log: this.log, item, isMatching });
+    return isMatching;
   }
 }
 
 export class MatchAll implements ItemFilter {
   private readonly filters: ItemFilter[];
+  private log: Logger;
 
-  constructor(filters: ItemFilter[]) {
+  constructor(filters: ItemFilter[], log: Logger = defaultLogger) {
     this.filters = filters;
+    this.log = log.child({ class: this.constructor.name });
   }
 
   async match(item: MatchableItem): Promise<boolean> {
@@ -54,15 +101,19 @@ export class MatchAll implements ItemFilter {
       this.filters.map((filter) => filter.match(item)),
     );
 
-    return results.every((result) => result);
+    const isMatching = results.every((result) => result);
+    logMatchResult({ log: this.log, item, isMatching });
+    return isMatching;
   }
 }
 
 export class MatchAny implements ItemFilter {
   private readonly filters: ItemFilter[];
+  private log: Logger;
 
-  constructor(filters: ItemFilter[]) {
+  constructor(filters: ItemFilter[], log: Logger = defaultLogger) {
     this.filters = filters;
+    this.log = log.child({ class: this.constructor.name });
   }
 
   async match(item: MatchableItem): Promise<boolean> {
@@ -70,7 +121,9 @@ export class MatchAny implements ItemFilter {
       this.filters.map((filter) => filter.match(item)),
     );
 
-    return results.some((result) => result);
+    const isMatching = results.some((result) => result);
+    logMatchResult({ log: this.log, item, isMatching });
+    return isMatching;
   }
 }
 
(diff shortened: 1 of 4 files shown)


src/system.ts
```diff
@@ -233,10 +233,13 @@ export const contiguousDataFsCacheCleanupWorker = !isNaN(
...
-  { name: 'Bundle-Format', value: 'binary' },
-  { name: 'Bundle-Version', valueStartsWith: '2.' },
-]);
+const ans104TxMatcher = new MatchTags(
+  [
+    { name: 'Bundle-Format', value: 'binary' },
+    { name: 'Bundle-Version', valueStartsWith: '2.' },
+  ],
+  log,
+);
```
djwhitt, line 242: When I run this on blocks with bundles, I get log messages like this:

```
debug: filter is not matching {"class":"MatchTags","id":"JyWSXVikQgG5bsrBRBCaVF2_URuns8QhVgIdoC5lbTk","timestamp":"2025-02-13T19:20:44.869Z"}
debug: filter is matching {"class":"MatchTags","id":"-27XiyI1WC0SIBDuLY0Qp7KTA6cGrNsbVRR1cm539_w","timestamp":"2025-02-13T19:20:44.869Z"}
debug: filter is matching {"class":"MatchTags","id":"x1bM4WIvj5XfAu3XNmbct560gq-vDKMIGK2uuBEa1Zw","timestamp":"2025-02-13T19:20:44.870Z"}
debug: filter is not matching {"class":"MatchTags","id":"M6WX8ac2ovjGYkTlI4mbw911z8hNd9wmWBJOGQzyvkM","timestamp":"2025-02-13T19:20:44.870Z"}
debug: filter is not matching {"class":"MatchTags","id":"MQW5iH3A7Wh4zO-3l1K3GFABfeKdVDzrKOvuZyuWL6U","timestamp":"2025-02-13T19:20:44.870Z"}
debug: filter is not matching {"class":"MatchTags","id":"WQcUzQkPHgF342XJbPYkekYeYZ-0fFW0ellnf4Vdy8A","timestamp":"2025-02-13T19:20:44.870Z"}
```

Something seems not quite right about how this is getting logged.

src/workers/ans104-data-indexer.ts
```diff
@@ -32,6 +32,7 @@ const DEFAULT_WORKER_COUNT = 1;
 
 export class Ans104DataIndexer {
   // Dependencies
+
```
djwhitt, line 35: Looks like some extra white space got added here.

src/log.ts
```diff
@@ -25,6 +25,25 @@ const LOG_ALL_STACKTRACES =
...
+// in the filters.ts file, we want to build a tree of filters
+// so we concatinate previous filter string to the next one
+const appendItemFilterFormat = format((info, opts: unknown) => {
+  if (
+    typeof opts === 'object' &&
+    typeof (opts as ItemFilterChildLoggerOpts).itemFilter === 'string'
+  ) {
+    const optsFromChildLogger = opts as ItemFilterChildLoggerOpts;
+    info.itemFilter = info.itemFilter
+      ? `${info.itemFilter} AND ${optsFromChildLogger.itemFilter}`
```
djwhitt, line 41: I was trying to figure out exactly what this does and looking for `AND` in the output and didn't see it. Do you have an example log message form you testing where this shows up? It's quite possible it's there and I'm just not seeing it.

src/filters.ts
```diff
@@ -191,27 +256,63 @@ export class MatchNestedBundle implements ItemFilter {
  *
  *   { never: true }
  */
-export function createFilter(filter: any): ItemFilter {
+export function createFilter(
+  filter: any,
+  logger: Logger = defaultLogger,
+): ItemFilter {
+  const log = logger.child({ class: 'ItemFilter' });
```
djwhitt, line 263: Let's use `function: 'createFilter'` instead.

djwhitt (commented):
Added round 1 comments.

src/system.ts
hlolli: correct, there's a second tag matcher instantiated, while I do pass the logger I am forgetting to give it the same treatment, fixing

hlolli: I make it consistent now, but we could imagine that debugging this matcher isn't really as important and we could consider a no-op logger (or not pass it at all)

src/log.ts
hlolli: first: this is silly code because if I pass itemFilter, it will have overridden and itemFilter previously in the logger context.
Secondly, I've been going back and forth on this. And we don't need this I think. All the "or" and "and" higherarchy is already captured in the map operators in createFilter and we pass logger to each one. So we are ok here.
I'm adding deeply nested "and" "or "tree" and I don't see any loss of contex. OTOH this isn't very plesantly to debug with the json strings compressed into 1 liners. But that's different topic.

So, I'm removing this logger filter for now.

src/filters.ts
hlolli: done

src/workers/ans104-data-indexer.ts
hlolli: removed

src/system.ts
hlolli: the screenshot version is on the branch now

src/filters.ts
```diff
@@ -17,60 +17,113 @@
...
+    });
+  }
+};
 
 export class AlwaysMatch implements ItemFilter {
-  async match(_: MatchableItem): Promise<boolean> {
+  private log: Logger;
+
+  constructor(log: Logger = defaultLogger) {
+    this.log = log.child({ class: 'AlwaysMatch' });
```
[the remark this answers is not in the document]
hlolli: good idea, fixed

src/filters.ts
```diff
@@ -17,60 +17,113 @@
...
+  isMatching: boolean;
+}) => {
+  if (isMatching) {
+    log.debug('filter is matching', {
+      id: item.id,
+      height: item.height,
+      parent: item.parent_id,
+    });
+  } else {
+    log.debug('filter is not matching', {
```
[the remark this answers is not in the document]
hlolli: done

src/filters.ts
```diff
@@ -17,60 +17,113 @@
...
+  log,
+  item,
+  isMatching,
+}: {
+  log: Logger;
+  item: MatchableItem;
+  isMatching: boolean;
+}) => {
+  if (isMatching) {
+    log.debug('filter is matching', {
```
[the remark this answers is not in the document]
hlolli: silly it is, and setting log level to silly does enable these

src/lib/ans-104.ts
```diff
@@ -410,7 +410,10 @@ if (!isMainThread) {
     });
   };
 
-  const filter = createFilter(JSON.parse(workerData.dataItemIndexFilterString));
+  const filter = createFilter(
+    JSON.parse(workerData.dataItemIndexFilterString),
+    log,
```
djwhitt, line 415: Do we need a child logger here as well?

src/filters.ts
```diff
@@ -191,27 +256,63 @@ export class MatchNestedBundle implements ItemFilter {
...
+      itemFilter: JSON.stringify({ not: filter.not }),
+    });
+    return new NegateMatch(createFilter(filter.not, childLogger), childLogger);
   } else if (filter?.and) {
-    return new MatchAll(filter.and.map(createFilter));
+    const childLogger = log.child({
+      itemFilter: JSON.stringify({ and: filter.and }),
+    });
+    return new MatchAll(
+      filter.and.map((and: any) => createFilter(and, childLogger)),
```
djwhitt, line 297: I think we could make this a bit nicer if we create a filter "path" that was passed along and included in the log context. The top level `itemFilterPath` would be `[]`. When a filter with other nested filters is create the filter at that level would be appended (be sure to create a copy of the array when doing this!). Then when each child logger is created, the current path + the filter in question is added. In the end you'd get paths like this `[and, or, or, tags]`.

djwhitt (commented):
I think there's another spot where we're missing log context. I also added a suggestion about how we could include filter paths in the log messages.

src/filters.ts
hlolli: I see what you mean, I'll implement it and we'll see how it turns out.

src/lib/ans-104.ts
hlolli: Good catch, I felt it was in the class, it's quite a global filter, I'm setting it to

```ts
log.child({ function: 'ans104IndexFilter' })
```

but any idea for context object could be good or better than this

I'd say Index filter because this is only used once as indexing filter

djwhitt (approved):

PR #326, opened 2025-02-13
refactor(logging): improve log message clarity for better debugging

hlolli:


src/workers/data-importer.ts
```diff
@@ -32,9 +32,9 @@ interface IndexProperty {
 }
 
 type AnyContiguousData = { id: string };
-type UnbundleableItem = (NormalizedDataItem | PartialJsonTransaction) &
+type TxWithParentIndex = (NormalizedDataItem | PartialJsonTransaction) &
```
djwhitt, line 35: I think the original name for this is fine. We can only unbundle items with indexes, and they can be either data items or transactions so we shouldn't include `Tx` in the name.

A comment explaining that `Ans104Parser` requires items with indexes would be helpful though for clarity.

src/workers/data-importer.ts
```diff
@@ -154,7 +154,7 @@ export class DataImporter {
     return this.queue.length() >= this.maxQueueSize;
   }
 
-  isUnbundleableItem(item: ImportableItem): item is UnbundleableItem {
+  txHasIndexProperty(item: ImportableItem): item is TxWithParentIndex {
```
djwhitt, line 157: This should just be `hasIndexPropery`. It's not necessarily a transaction.

src/workers/data-importer.ts
```diff
@@ -154,7 +154,7 @@ export class DataImporter {
...
     return Object.keys(item).length > 1 && 'index' in item;
   }
```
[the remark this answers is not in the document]
hlolli: without an 'id' it would be broken

src/workers/data-importer.ts
hlolli: changed

src/workers/data-importer.ts
hlolli: reverted with comments as mentioned

djwhitt (approved):

PR #328, opened 2025-02-17
chore(ao): increase max memory for AO CU

dtfiedler:
The ARIO process will require 8GiB of memory. This bumps the default memory supported by the sidecar AO CU to 8GiB.


docker-compose.ao.yaml
```diff
@@ -1,7 +1,7 @@
 services:
   ao-cu:
     image: ghcr.io/permaweb/ao-cu:latest
-    restart: on-failure
+    restart: unless-stopped
```
dtfiedler, line 4: I now understand the issue originally reported - the CU will continue to restart on SIGKILL - even if intentionally stopped. This blocks other docker containers from starting up when necessary (e.g. in our e2e of the ar-io-sdk and ar-io-network-process tests that use their own CUs)

djwhitt (approved):

karlprieb (approved):

PR #329, opened 2025-02-18
tests: add ability to tag tests and tag flaky tests PE-7696

djwhitt:
This adds an `isTestFiltered` function takes an array of tags and checks them against the comma separated lists in TEST_ONLY_TAGS and TEST_SKIP_TAGS. When TEST_ONLY_TAGS is used, only tests with tags in that list will be run. When TEST_SKIP_TAGS is used, any test with tag in that list will be skipped.

This change also tags tests that are currently flaky and adjusts the GitHub workflow so that they will be skipped.


test/end-to-end/arns.test.ts
```diff
@@ -24,6 +24,7 @@ import {
 } from 'testcontainers';
 import axios from 'axios';
 import { rimraf } from 'rimraf';
+import { isTestFiltered } from '../utils.js';
```
karlprieb, line 27: Why import it here if it's not being used?

karlprieb (approved):

test/end-to-end/arns.test.ts
djwhitt: Good catch. I think I added it initially before realizing we didn't need to skip the test in this file.

PR #330, opened 2025-02-18
test(e2e) avoid time based waits

karlprieb:
- removes all time based waits (only one left with a comment)
- use `START_WRITES: false` whenever we can so we don't need to import unnecessary blocks
- fix some tests after creating utility functions like:
  - waitForBlocks
  - waitForLogMessage
  - waitForBundleToBeIndexed
  - waitForTxToBeIndexed
  - waitForDataItemToBeIndexed
- reduce pooling interval to 1 second


test/end-to-end/arns.test.ts
```diff
@@ -120,7 +107,7 @@ describe('ArNS', function () {
         assert.ok(res.headers['x-arns-record-index'] === '0');
       });
 
-      it.skip('Verifying "ardrive.ar-io.localhost" X-ArNS-Undername-Limit header', async function () {
+      it('Verifying "ardrive.ar-io.localhost" X-ArNS-Undername-Limit header', async function () {
```
karlprieb, line 110: @djwhitt I'm not skipping this anymore as it was skipped by mistake

test/end-to-end/data-sources.test.ts
karlprieb, line 1: @djwhitt this one was fixed so not skipping it anymore

test/end-to-end/data.test.ts
```diff
@@ -241,24 +239,16 @@ describe('Data', function () {
   });
 });
 
-describe('X-Cache header', function () {
+describe('X-Cache header', { skip: isTestFiltered(['flaky']) }, function () {
```
karlprieb, line 242: @djwhitt skipping these 2 as they're flaky

test/end-to-end/data.test.ts
```diff
@@ -326,8 +308,7 @@ describe('X-AR-IO-Root-Transaction-Id header', function () {
     assert.equal(bundleRes.headers['x-ar-io-root-transaction-id'], undefined);
   });
 
-  // TODO: reenable when we have an endpoint that lets us wait for unbundling
-  it.skip('Verifying header for data item', async function () {
+  it('Verifying header for data item', async function () {
```
karlprieb, line 311: @djwhitt working now

test/end-to-end/indexing.test.ts
```diff
@@ -494,6 +465,12 @@ describe('Indexing', function () {
         },
       });
 
+      await waitForLogMessage({
+        container: compose.getContainer('core-1'),
+        expectedMessage: 'Transaction imported.',
+      });
+
+      // wait 10 seconds after transaction was imported to give it time to be indexed  (not be indexed in this case)
```
karlprieb, line 473: @djwhitt Not exactly sure what to do here to make sure tx was imported and bundle not indexed...

test/end-to-end/indexing.test.ts
```diff
@@ -916,37 +899,37 @@ describe('Indexing', function () {
...
     });
 
     after(async function () {
       await compose.down();
     });
 
-    it.skip('Verifying if data item headers were indexed', async function () {
-      const stmt = bundlesDb.prepare('SELECT * FROM new_data_items');
-      const dataItems = stmt.all();
+    it('Verifying if data item headers were indexed', async function () {
```
karlprieb, line 909: @djwhitt this one was fixed

test/end-to-end/metrics.test.ts
```diff
@@ -112,54 +107,42 @@ describe('Metrics', function () {
...
+      await cleanDb();
+
+      compose = await composeUp({
+        START_HEIGHT: START_HEIGHT.toString(),
+        STOP_HEIGHT: STOP_HEIGHT.toString(),
+      });
 
       coreDb = new Sqlite(`${projectRootPath}/data/sqlite/core.db`);
-      await waitForBlocks(coreDb, STOP_HEIGHT);
+      await waitForBlocks({ coreDb, stopHeight: STOP_HEIGHT });
```
karlprieb, line 118: @djwhitt using `waitForBlocks` fix this flaky test

test/end-to-end/webhook.test.ts
```diff
@@ -28,10 +27,12 @@ import {
 import { createServer } from 'node:http';
 import wait from 'wait';
 import axios from 'axios';
+import { cleanDb } from './utils.js';
+import { isTestFiltered } from '../utils.js';
 
 const projectRootPath = process.cwd();
 
-describe.skip('WebhookEmitter', () => {
+describe('WebhookEmitter', { skip: isTestFiltered(['flaky']) }, () => {
```
karlprieb, line 35: @djwhitt works in my machine :) but I know it's breaking when running through ci. I'll check old runs to understand why

test/end-to-end/utils.ts
```diff
@@ -15,16 +15,293 @@
...
+  host?: string;
+  secret?: string;
+  timeout?: number;
+  checkInterval?: number;
+}) => {
+  return waitFor({
+    check: () => getBundleStatus({ id, host, secret }),
+    validate: (bundleStatus) =>
+      bundleStatus?.lastUnbundledAt !== null &&
+      bundleStatus?.dataItemCount !== null,
```
djwhitt, line 260: I think the configuration option to reduce the interval and enable using it is on another branch right now, but we might want to wait for `lastFullyIndexedAt` here.

djwhitt (approved):
Very nice work! Just added one comment about the bundle indexing wait condition.

test/end-to-end/utils.ts
karlprieb: lastFullyIndexedAt is just updated when update bundle timestamps run. Unless we change the update bundle timestamps to something like 1 seconds we will not have lastFullyIndexedAt populated. Maybe we should do that but I thought that lastUnbundledAt and dataItemCount would be enough.
What do you think about that?

test/end-to-end/metrics.test.ts
```diff
@@ -104,7 +99,7 @@ async function getMetrics(): Promise<ParsedMetrics | undefined> {
 
 const projectRootPath = process.cwd();
 
-describe('Metrics', function () {
+describe('Metrics', { skip: isTestFiltered(['flaky']) }, function () {
```
karlprieb, line 102: @djwhitt skipping this one. Tests were passing because they were not asserting anything with the if blocks.

test/end-to-end/indexing.test.ts
djwhitt: Eh, we can leave as is for now. We might tag this one on in the future though.

djwhitt (approved):

PR #332, opened 2025-02-20
feat: update grafana dashboard json, make dynamic datasource

onekill0503:
The dashboard was defined with static datasource , so it's you need to modify it manually if ur grafana was installed in different server. I've add new panel "Datasource" for user select their datasource, also make the other panel integrated with the datasource panel.

### DIFF
--- a/monitoring/grafana/dashboards/default.json
+++ b/monitoring/grafana/dashboards/default.json
@@ -22,10 +22,7 @@
   "liveNow": false,
   "panels": [
     {
-      "datasource": {
-        "type": "prometheus",
-        "uid": "prometheus"
-      },
+      "datasource": "${datasource}",
       "fieldConfig": {
         "defaults": {
           "color": {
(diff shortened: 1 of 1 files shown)


hlolli:
these dashboards we have internally made with jsonnet/graphonnet. There's a chance we'll rather take that approach to dynamicism rather than javascript string manipulation of json. Let's see what other say :)

thanks for your PR! 🙏

onekill0503:
> these dashboards we have internally made with jsonnet/graphonnet. There's a chance we'll rather take that approach to dynamicism rather than javascript string manipulation of json. Let's see what other say :)
> 
> thanks for your PR! 🙏

your welcome :) , just want to share what I faced before. It was cool if there's efficient way.

dtfiedler (approved):
i've tested and it looks good. thanks @onekill0503

karlprieb (approved):

PR #338, opened 2025-02-26
feat(standalone-sqlite): retry saveBundle when SQLITE_BUSY

karlprieb:


src/database/standalone-sqlite.ts
```diff
@@ -2855,7 +2849,10 @@ export class StandaloneSqliteDatabase
     role: WorkerRoleName,
     method: WorkerMethodName,
     args: any,
+    retryIfFailed = false,
```
djwhitt, line 2852: What do you think about adjusting this to be a retry count that defaults to `0` instead of a boolean?

djwhitt (commented):
Looks good. Just added one suggestion/question about passing the retry count instead of a boolean.

src/database/standalone-sqlite.ts
karlprieb: [`7b8b3a8` (#338)](https://github.com/ar-io/ar-io-node/pull/338/commits/7b8b3a8258d604d6154d7d68ed5214c15b47aec3)

djwhitt (approved):

PR #339, opened 2025-02-27
fix(graphql): correctly handle empty cursor for clickhouse

karlprieb:
No tests were added as we currently doesn't have clickhouse tests


thiagocarvalhodev (approved):
LGTM

djwhitt (approved):
Yeah, lack of tests in on this is fine for now. We'll find out fast if this is still a problem for apps.

PR #341, opened 2025-03-04
fix(parquet): fix export parquet names

karlprieb:
Exported parquet file names are now aligned with what import parquet script expects.

Example:
```
blocks-0-185-rowCount:186.parquet
transactions-0-185-rowCount:331.parquet
```

### DIFF
--- a/src/workers/parquet-exporter.ts
+++ b/src/workers/parquet-exporter.ts
@@ -58,6 +58,12 @@ type ExportData = {
   error?: string;
 };
 
+type HeightRange = {
+  startHeight: bigint;
+  endHeight: bigint;
+  rowCount: bigint;
+};
+
 export class ParquetExporter {
   private log: winston.Logger;
   private worker: Worker | null = null;
@@ -94,7 +100,7 @@ export class ParquetExporter {
     outputDir: string;
     startHeight: number;
     endHeight: number;
-    maxFileRows: number;
+    maxFileRows?: number;
   }): Promise<void> {
     if (this.exportStatus.status === RUNNING) {
       const error = new Error('An export is already in progress');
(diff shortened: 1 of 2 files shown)


src/workers/parquet-exporter.ts
```diff
@@ -488,34 +494,45 @@ const exportToParquet = async ({
...
+      (maxFileRows !== undefined && rowCount >= maxFileRows) ||
+      height === maxHeight
+    ) {
+      heightRanges.push({
+        startHeight: currentRangeStart,
+        endHeight: height,
+        rowCount,
+      });
+
+      const fileName = `${tableName}-${currentRangeStart}-${height}-rowCount:${rowCount}.parquet`;
```
djwhitt, line 524: We should keep the `minHeight:` and `maxHeight:` prefixes if possible. Did you run into an issue importing with those? I think it should work okay.

djwhitt (commented):
One comment/question about filenames. Otherwise this looks good.

src/workers/parquet-exporter.ts
karlprieb: It will works as `sed 's/blocks-//;s/-rowCount:[0-9]*\.parquet//'` from https://github.com/ar-io/ar-io-node/blob/b30e8d2ba9d7794c31a9cb30ef23dbe68620b3c6/scripts/clickhouse-import#L9 will get anything between `blocks-` and `-rowCount`, but I can add `minHeight:` and `maxHeight:` to keep it more readable

djwhitt: Yeah, I'd like to keep them so we can more easily parse them out of the file name in the future.

djwhitt (approved):

PR #344, opened 2025-03-11
feat(router): serve tx or arns name from apex domain

karlprieb:


src/app.ts
```diff
@@ -30,6 +31,11 @@ import { apolloServer } from './routes/graphql/index.js';
 import { openApiRouter } from './routes/openapi.js';
 import * as system from './system.js';
 
+if (config.APEX_TX_ID !== undefined && config.APEX_ARNS_NAME !== undefined) {
+  log.error('Cannot set both APEX_TX_ID and APEX_ARNS_NAME');
```
hlolli, line 35: nit:
I'm thinking if it's immediately obvious or not that it's a user configuration error, could be written: Please only specify env-var APEX_TX_ID or env-var APEX_ARNS_NAME but not both.

src/routes/default.ts
```diff
@@ -0,0 +1,55 @@
...
+  const modifiedReq = Object.create(req);
+
+  if (APEX_TX_ID !== undefined) {
+    modifiedReq.params = { 0: APEX_TX_ID, ...req.params };
+    return dataHandler(modifiedReq, res, next);
+  }
+
+  if (APEX_ARNS_NAME !== undefined) {
+    // Use Proxy to modify hostname as it is read only
+    const modifiedReq = new Proxy(req, {
```
hlolli, line 41: modifiedReq is already in scope, declared as const, is this a legal operation?

src/routes/default.ts
```diff
@@ -0,0 +1,55 @@
...
+import { arIoInfoHandler } from './ar-io.js';
+
+export const defaultRouter = Router();
+
+defaultRouter.get('/', async (req, res, next) => {
+  if (APEX_TX_ID === undefined && APEX_ARNS_NAME === undefined) {
+    return arIoInfoHandler(req, res);
+  }
+
+  const modifiedReq = Object.create(req);
```
hlolli, line 32: Could you write a comment here about why Object.create and not spread operator or Object.clone or Object.copy ?

src/routes/default.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  if (APEX_TX_ID !== undefined) {
+    modifiedReq.params = { 0: APEX_TX_ID, ...req.params };
+    return dataHandler(modifiedReq, res, next);
+  }
+
+  if (APEX_ARNS_NAME !== undefined) {
+    // Use Proxy to modify hostname as it is read only
+    const modifiedReq = new Proxy(req, {
+      get: (target, prop) => {
```
hlolli, line 42: I enjoy seeing Proxy being used in a practical way. The alternative to using Proxy would be to patch all the req.hostname and req.subdomains getters ?

src/routes/default.ts
```diff
@@ -0,0 +1,55 @@
...
+        }
+        if (prop === 'subdomains') {
+          return [APEX_ARNS_NAME];
+        }
+        return target[prop as keyof typeof target];
+      },
+    });
+
+    return arnsMiddleware(modifiedReq, res, next);
+  }
```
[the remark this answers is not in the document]
djwhitt: @karlprieb I agree with this. We should log an error in `config.ts` if it's not set.

src/routes/default.ts
djwhitt: Yes. The `if` will creates a new scope. That said, I think the earlier declaration could be moved inside the `APEX_TX_ID` conditional for clarity.

src/app.ts
djwhitt: I agree with the general suggestion, but I would go with "APEX_TX_ID and APEX_ARNS_NAME are mutually exclusive but both are set." as the message.

djwhitt (commented):
@karlprieb Functionality looks fine, but I agree with the comments about error messages - both improving existing message and adding one.

src/routes/default.ts
karlprieb: I'll change to a spread operator as we use it more often. No particular reason to use `Object.create`

src/routes/default.ts
karlprieb: There's no safe alternatives as `hostname` is "read-only". `subdomains` are not though

hlolli: asking for my own understanding, feel free to ignore.
There's no "set"/setter implemented, so the read-only doesn't matter for the proxy?

src/app.ts
karlprieb: [`056e8c0` (#344)](https://github.com/ar-io/ar-io-node/pull/344/commits/056e8c088e7971abb2d68ef117ca6d21de4939c6)

src/routes/default.ts
karlprieb: [`0312eea` (#344)](https://github.com/ar-io/ar-io-node/pull/344/commits/0312eea635037077707b3401c106c7767a17b45f)

src/routes/default.ts
karlprieb: I kept `Object.clone` as I was having a hard time making TS happy with the new `Request` when using spread operator

djwhitt (approved):

hlolli (approved):

PR #353, opened 2025-03-14
chore(grafana): add default dashboard for AO cu

dtfiedler:
image


karlprieb:
@dtfiedler Can you use a dynamic dataset as #332 is doing?

dtfiedler:
> @dtfiedler Can you use a dynamic dataset as #332 is doing?

yep - will update!

src/resolution/composite-arns-resolver.ts
```diff
@@ -143,7 +143,7 @@ export class CompositeArNSResolver implements NameResolver {
             }
           } catch (error: any) {
             this.log.error('Error resolving name with resolver', {
-              resolver,
+              resolver: resolver.constructor.name,
```
dtfiedler, line 146: this wasn't getting logged

dtfiedler:
Included PE-7839

PR #356, opened 2025-03-18
feat(graphql): owner resolvers for clickhouse

karlprieb:
Huge commit including several refactors and the new OwnerFetcher.

Added:
- return owner_size and owner_offset from clickhouse
- generic AttributeFetcher
- OwnerFetcher, resolves owner key from offsets or fetch from chain
- generic B64UrlStore
- OwnerStore
- generic JsonStore
- DataItemAttributesStore
- TransactionAttributesStore

Changed:
- SignatureFetcher based on AttributeFetcher
- SignatureStore


src/data/attribute-fetcher.ts
```diff
@@ -0,0 +1,418 @@
...
+    });
+
+    let buffer = Buffer.alloc(0);
+
+    for await (const chunk of stream) {
+      buffer = Buffer.concat([buffer, chunk]);
+    }
+
+    return toB64Url(buffer);
+  }
```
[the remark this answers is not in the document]
hlolli: I see that we are locked into Buffer type, so, this comment doesn't matter

hlolli (approved):

src/data/attribute-fetchers.ts
```diff
@@ -0,0 +1,428 @@
...
+      buffer = Buffer.concat([buffer, chunk]);
+    }
+
+    return toB64Url(buffer);
+  }
+
+  async getDataItemAttributes(
+    id: string,
+  ): Promise<DataItemAttributes | undefined> {
+    const log = this.log.child({ method: 'getDataItemAttribute' });
```
djwhitt, line 95: Small typo here. Should be `getDataItemAttributes` (with an 's').

src/data/attribute-fetchers.ts
```diff
@@ -0,0 +1,428 @@
...
+        return transactionAttributes.owner;
+      }
+
+      this.log.warn('No attributes found for transaction', { id });
+
+      let ownerFromChain;
+
+      const ownerChainField = await this.chainSource.getTxField(id, 'owner');
+
+      if (
```
djwhitt, line 401: Worth adding a comment explaining why we do this.

src/types.d.ts
```diff
@@ -690,8 +709,3 @@ export type KVBufferStore = {
   has(key: string): Promise<boolean>;
   close(): Promise<void>;
 };
-
-export interface SignatureSource {
```
djwhitt, line 694: Even though we don't require them, I'd like to keep a `SignatureSource` and `OwnerSource` interface here for documentation and also implement them on the fetchers. That makes it clear what we need to implement if we wanted to pull these from other sources in the future.

djwhitt (commented):
Code and local testing look good. Added nits about a typo, comments, and interfaces.

src/data/attribute-fetchers.ts
karlprieb: https://github.com/ar-io/ar-io-node/pull/356/commits/15e95b6af5daae18b1af19c74a1f6ced5d729a51

src/types.d.ts
karlprieb: https://github.com/ar-io/ar-io-node/pull/356/commits/381b33e48a393a50a48131067507ad05fdd16ba2

karlprieb: @djwhitt Sorry! I read it wrong.
Here's the update [`381b33e` (#356)](https://github.com/ar-io/ar-io-node/pull/356/commits/381b33e48a393a50a48131067507ad05fdd16ba2)

djwhitt (approved):

PR #357, opened 2025-03-18
ci: trigger envoy deployment after image build

hlolli:


djwhitt (commented):
Does this handle updating the prod deployment when there's a merge to `main`?

dtfiedler (approved):

PR #359, opened 2025-03-18
fix(arns): add circuit breaker for fetching arns names

dtfiedler:
This adds a circuit breaker when fetching arns name list from a CU  with default failure rate of 30% and backoff period of 10 minutes. Doing so should prevent overwhelming CUs and rely more heavily on the  internal cache of the arns name list during the backoff period.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -586,6 +586,29 @@ export const ARNS_NAMES_CACHE_TTL_SECONDS = +env.varOrDefault(
   `${60 * 60}`, // 1 hour
 );
 
+export const ARNS_NAMES_CACHE_CIRCUIT_BREAKER_TIMEOUT_MS = +env.varOrDefault(
+  'ARNS_NAMES_CACHE_CIRCUIT_BREAKER_TIMEOUT_MS',
+  `${10 * 1000}`, // 10 seconds
+);
+
+export const ARNS_NAMES_CACHE_CIRCUIT_BREAKER_ERROR_THRESHOLD_PERCENTAGE =
+  +env.varOrDefault(
+    'ARNS_NAMES_CACHE_CIRCUIT_BREAKER_ERROR_THRESHOLD_PERCENTAGE',
+    '30', // 30% failure limit before circuit breaker opens
+  );
+
+export const ARNS_NAMES_CACHE_CIRCUIT_BREAKER_ROLLING_COUNT_TIMEOUT_MS =
+  +env.varOrDefault(
+    'ARNS_NAMES_CACHE_CIRCUIT_BREAKER_ROLLING_COUNT_TIMEOUT_MS',
+    `${60 * 1000}`, // 1 minute
+  );
+
+export const ARNS_NAMES_CACHE_CIRCUIT_BREAKER_RESET_TIMEOUT_MS =
+  +env.varOrDefault(
+    'ARNS_NAMES_CACHE_CIRCUIT_BREAKER_RESET_TIMEOUT_MS',
+    `${10 * 60 * 1000}`, // 10 minutes
+  );
+
 export const ARNS_NAME_LIST_CACHE_MISS_REFRESH_INTERVAL_SECONDS =
   +env.varOrDefault(
     'ARNS_NAME_LIST_CACHE_MISS_REFRESH_INTERVAL_SECONDS',
--- a/src/resolution/arns-names-cache.ts
+++ b/src/resolution/arns-names-cache.ts
@@ -30,6 +30,8 @@ import { connect } from '@permaweb/aoconnect';
 import { KvDebounceStore } from '../store/kv-debounce-store.js';
 import { KVBufferStore } from '../types.js';
 import * as metrics from '../metrics.js';
+import CircuitBreaker from 'opossum';
+
 const DEFAULT_CACHE_MISS_DEBOUNCE_TTL =
   config.ARNS_NAME_LIST_CACHE_MISS_REFRESH_INTERVAL_SECONDS * 1000;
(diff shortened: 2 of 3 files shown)


src/config.ts
```diff
@@ -586,6 +586,29 @@ export const ARNS_NAMES_CACHE_TTL_SECONDS = +env.varOrDefault(
   `${60 * 60}`, // 1 hour
 );
 
+export const ARNS_NAMES_CACHE_CIRCUIT_BREAKER_TIMEOUT_MS = +env.varOrDefault(
+  'ARNS_NAMES_CACHE_CIRCUIT_BREAKER_TIMEOUT_MS',
+  `${10 * 1000}`, // 10 seconds
```
dtfiedler, line 591: 10 seconds per records request is more than reasonable

src/config.ts
```diff
@@ -586,6 +586,29 @@ export const ARNS_NAMES_CACHE_TTL_SECONDS = +env.varOrDefault(
...
+);
+
+export const ARNS_NAMES_CACHE_CIRCUIT_BREAKER_ERROR_THRESHOLD_PERCENTAGE =
+  +env.varOrDefault(
+    'ARNS_NAMES_CACHE_CIRCUIT_BREAKER_ERROR_THRESHOLD_PERCENTAGE',
+    '30', // 30% failure limit before circuit breaker opens
```
dtfiedler, line 597: i.e 3/10 requests fail - we trigger the circuit breaker

src/config.ts
```diff
@@ -586,6 +586,29 @@ export const ARNS_NAMES_CACHE_TTL_SECONDS = +env.varOrDefault(
...
+export const ARNS_NAMES_CACHE_CIRCUIT_BREAKER_ROLLING_COUNT_TIMEOUT_MS =
+  +env.varOrDefault(
+    'ARNS_NAMES_CACHE_CIRCUIT_BREAKER_ROLLING_COUNT_TIMEOUT_MS',
+    `${60 * 1000}`, // 1 minute
+  );
+
+export const ARNS_NAMES_CACHE_CIRCUIT_BREAKER_RESET_TIMEOUT_MS =
+  +env.varOrDefault(
+    'ARNS_NAMES_CACHE_CIRCUIT_BREAKER_RESET_TIMEOUT_MS',
+    `${10 * 60 * 1000}`, // 10 minutes
```
dtfiedler, line 609: wait 10 minutes before we hit the CU again

src/config.ts
djwhitt: Did you test what we see normally?

src/config.ts
djwhitt: I was wondering if we should make the window smaller, but seems fine since 3 consecutive failures over 6 minutes could trigger this.

djwhitt (commented):
Code looks good, but please add comments above each environment variable in `config.ts` so we can start using that as our source of truth to generate docs.

Also, the environment variables need to be added to `docker-compose.yaml`.

src/config.ts
dtfiedler: i'm going off of grafana stats for cu.ardrive.io - the p99 there is < 2 seconds for a dry-run. will do similar tests against FWD CUs and add a comment

dtfiedler:
@djwhitt - added environment variable descriptions in [6e65f4e](https://github.com/ar-io/ar-io-node/pull/359/commits/6e65f4eb024d1a4478212d88b171ec2ff6ba1e15)

before i update docker compose - are you okay with peer list & arns name cache using the same circuit breaker environment variables like in https://github.com/ar-io/ar-io-node/pull/360 - or would you prefer them separated?

djwhitt:
@dtfiedler Same is fine. The peers list is less critical, so stricter timing is unlikely to hurt anything there.

src/config.ts
```diff
@@ -586,6 +586,42 @@ export const ARNS_NAMES_CACHE_TTL_SECONDS = +env.varOrDefault(
...
+export const ARIO_PROCESS_DEFAULT_CIRCUIT_BREAKER_ERROR_THRESHOLD_PERCENTAGE =
+  +env.varOrDefault(
+    'ARIO_PROCESS_DEFAULT_CIRCUIT_BREAKER_ERROR_THRESHOLD_PERCENTAGE',
+    '30', // 30% failure limit before circuit breaker opens
+  );
+
+// Defines the time window for tracking errors when retrieving ARIO process state from AO
+// The circuit breaker counts failures within this rolling time window to determine
+// if the error threshold percentage has been exceeded
+export const ARIO_PROCESS_DEFAULT_CIRCUIT_BREAKER_ROLLING_COUNT_TIMEOUT_MS =
```
djwhitt, line 610: Is this enough time if we're waiting 2 minutes between refreshes?

dtfiedler: good call out -

if increased (i.e., adjusted to something like 10 minutes), this circuit breaker would trigger if 2/5 (40%) refreshes fail. the circuit breaker would then prevent requests to the CU for 10 minutes. 

seems reasonable enough? we may want to consider a longer backoff period (~20 minutes) as well - given CU issues tend to linger.

djwhitt: Yeah, 2/5 seems good.

djwhitt: I'm fine with a 20 minute back-off as well. Once we implement the better trusted gateway fallback that should be quite safe.

dtfiedler: updated to 10 min window and 20 minute back-off in [fc0e72d](https://github.com/ar-io/ar-io-node/pull/359/commits/fc0e72d62f5845c001f7ab7f6141fa2ab0ed320e)

djwhitt (approved):

PR #361, opened 2025-03-20
feat(chunk-data-store): reduce log verbosity when  chunk is already cached

hlolli:


src/store/fs-chunk-data-store.ts
```diff
@@ -115,10 +115,23 @@ export class FsChunkDataStore implements ChunkDataStore {
         `${process.cwd()}/${chunkHashPath}`,
       );
 
-      await fs.promises.symlink(
-        targetPath,
-        this.chunkDataRootPath(dataRoot, relativeOffset),
-      );
+      const sourcePath = this.chunkDataRootPath(dataRoot, relativeOffset);
+
+      if (fs.existsSync(sourcePath)) {
```
djwhitt, line 120: This looks backwards to me. Wouldn't we want to unlink the target? Also, why do anything in this case. If the symlink is already there we shouldn't need to replace it.

hlolli: now modified to just reduce the log verbosity undir this somewhat common and harmless condition

src/store/fs-chunk-data-store.ts
```diff
@@ -115,14 +115,27 @@ export class FsChunkDataStore implements ChunkDataStore {
...
 
-      await fs.promises.symlink(
-        targetPath,
-        this.chunkDataRootPath(dataRoot, relativeOffset),
-      );
-      this.log.info('Successfully cached chunk data', {
-        dataRoot,
-        relativeOffset,
-      });
+      const sourcePath = this.chunkDataRootPath(dataRoot, relativeOffset);
```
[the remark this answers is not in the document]
hlolli: done

djwhitt (approved):

djwhitt (approved):

PR #364, opened 2025-03-21
ci(deploy-core-image): fix main branch trigger condition

hlolli:


.github/workflows/deploy-core-image.yml
```diff
@@ -4,14 +4,18 @@ on:
...
+      - main
 
 jobs:
   ar-io-core-deployment:
     runs-on: ubuntu-latest
     if: >
-      github.event.workflow_run.conclusion == 'success' &&
-      (github.event.workflow_run.head_branch == 'develop' ||
-       github.event.workflow_run.head_branch == 'main')
+      github.event_name == 'push' ||
```
djwhitt, line 15: I'd feel better if we included a `github.event.workflow_run.head_branch == 'main'` condition here too.

hlolli: fixed

djwhitt (approved):

PR #368, opened 2025-04-01
feat(filters): make filters sync in order to make them usable for log-filters

hlolli:


djwhitt (commented):
The `async` is intentional. There's no guarantee that all future matchers will be synchronous.

djwhitt (approved):
Discussed on Slack. Apply filters to logs requires this. If we need to do anything async for filters we can do it when they're initialized.

PR #371, opened 2025-04-07
feat: s3 chunk data source

hlolli:
feat: add S3-based chunk data source and unify getChunkDataByAny signature

- Adds a new `S3ChunkSource` implementation of `ChunkDataByAnySource`, enabling
  chunk retrieval from S3 using the Arweave-compatible `/chunk/<offset>` interface.
- Refactors `getChunkDataByAny` and related methods to use a unified
  `ChunkDataByAnySourceParams` object for improved consistency and readability.

### DIFF
--- a/src/data/s3-chunk-source.ts
+++ b/src/data/s3-chunk-source.ts
@@ -1,0 +1,84 @@
+/**
+ * AR.IO Gateway
+ * Copyright (C) 2022-2023 Permanent Data Solutions, Inc. All Rights Reserved.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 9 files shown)


src/data/s3-chunk-source.ts
```diff
@@ -0,0 +1,93 @@
...
+    dataRoot,
+    relativeOffset,
+  }: ChunkDataByAnySourceParams): Promise<ChunkData> {
+    // Validate that dataRoot and relativeOffset exist
+    if (!dataRoot || relativeOffset == null) {
+      throw new Error(
+        'S3ChunkSource.getChunkDataByAny called without dataRoot or relativeOffset',
+      );
+    }
+    // Build the S3 object key: /<prefix>/<dataRoot>/<relativeOffset>
```
djwhitt, line 65: I think you can strip out some of these obvious comments. :slightly_smiling_face:

src/types.d.ts
```diff
@@ -481,31 +488,23 @@ export interface Chunk extends ChunkMetadata, ChunkData {
   tx_path: Buffer;
 }
 
+export interface ChunkDataByAnySourceParams {
```
djwhitt, line 491: I'm thinking we should make this a union type of some sort. We either need absolute offset and size or size, data root, and relative offset. That could be something we revisit at the end though when we're putting all the pieces together.

src/data/s3-chunk-source.ts
hlolli: cleaned up

PR #372, opened 2025-04-09
feat: add robust TX header fetching with multiple weighted peers

hlolli:
- Replace `weightedChunkPeers` references with a single `weightedPeers` array to standardize weighted peer handling.
- Rename `selectChunkPeers()` to `selectPeers()` for broader usability.
- Update `peerGetTx()` to:
  - Attempt fetching from multiple weighted peers before failing over to the trusted node.
  - Adjust peer weights on success or failure using `handlePeerSuccess()` and `handlePeerFailure()`.
- Switch chunk-fetch logic calls from `selectChunkPeers` to `selectPeers`.
- Ensure peer weighting logic applies consistently for both chunk and transaction fetching.
- Increase the default peer request timeout in `peerGetTx()` from 500ms to 1000ms.

### DIFF
--- a/src/arweave/composite-client.ts
+++ b/src/arweave/composite-client.ts
@@ -82,6 +82,8 @@ interface Peer {
   lastSeen: number;
 }
 
+type WeightedPeerListName = 'weightedChainPeers' | 'weightedChunkPeers';
+
 export class ArweaveCompositeClient
   implements
     ChainSource,
@@ -128,6 +130,7 @@ export class ArweaveCompositeClient
   // Peers
   private peers: Record<string, Peer> = {};
   private preferredPeers: Set<Peer> = new Set();
+  private weightedChainPeers: WeightedElement<string>[] = [];
   private weightedChunkPeers: WeightedElement<string>[] = [];
 
   // Block and TX promise caches used for prefetching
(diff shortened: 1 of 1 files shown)


src/arweave/composite-client.ts
```diff
@@ -441,9 +441,9 @@ export class ArweaveCompositeClient
 
             // Prefetch transactions
             if (prefetchTxs) {
-              block.txs.forEach(async (txId: string) => {
-                this.prefetchTx({ txId });
-              });
+              for (const txId of block.txs) {
+                await this.prefetchTx({ txId });
```
djwhitt, line 445: The lack of `await` here was intentional. We want to "fire and forget" all the TX prefetches when a block is retrieved. The promises get awaited later then `getTx` is called.

src/arweave/composite-client.ts
```diff
@@ -554,62 +565,62 @@ export class ArweaveCompositeClient
...
+        return response.data;
+      } catch (error) {
+        // If something goes wrong, remove it from the store (in case partially cached)
+        this.txStore.del(txId);
         this.log.warn('Transaction prefetch failed:', {
-          txId: txId,
-          message: error.message,
-          stack: error.stack,
+          txId,
+          message: (error as Error).message,
```
djwhitt, line 613: Nit: but can we push this `as` (or similar) up into the earlier `catch` line?

src/arweave/composite-client.ts
```diff
@@ -554,62 +565,62 @@ export class ArweaveCompositeClient
     const url = `/${transactionType}/${txId}`;
     let downloadedFromPeer = true;
 
-    const responsePromise = this.txStore
-      .get(txId)
-      .then((tx) => {
+    const responsePromise = (async () => {
```
djwhitt, line 568: Nice refactor! :clap:

src/arweave/composite-client.ts
```diff
@@ -364,16 +364,16 @@ export class ArweaveCompositeClient
     }
   }
 
-  selectChunkPeers(peerCount: number): string[] {
-    const log = this.log.child({ method: 'selectChunkPeers' });
+  selectPeers(peerCount: number): string[] {
```
djwhitt, line 367: We should use a separate weights for chunks and headers. I expect headers to be much more widely available than chunks. We don't want header availability affecting chunk weights.

src/arweave/composite-client.ts
```diff
@@ -514,27 +514,38 @@ export class ArweaveCompositeClient
...
-      method: 'GET',
-      url,
-      baseURL: randomPeer.url,
-      timeout: 500,
-    }).then(async (response) => {
-      const tx = this.arweave.transactions.fromRaw(response.data);
-      const isValid = await this.arweave.transactions.verify(tx);
-      if (!isValid) {
-        throw new Error('Invalid peer fetched transaction');
+    for (const peerUrl of peersToTry) {
```
djwhitt, line 520: I think we should use `Promise.any` here since TX header fetching is somewhat performance sensitive.

src/arweave/composite-client.ts
hlolli: I see, I felt it wasn't intentional and was an extra layer of flooding. Yeh I'll remove the await and add comment about this.

src/arweave/composite-client.ts
hlolli: in the catch clause itself: `Catch clause variable type annotation must be 'any' or 'unknown' if specified.ts(1196)` I'll reassign it to a new variable to cast it to Error.

src/arweave/composite-client.ts
hlolli: done, I solved this by passing a string matching the class property name of the weighter peers list

src/arweave/composite-client.ts
hlolli: more if that async IIF pattern, ugly wrapper but good profits

src/arweave/composite-client.ts
hlolli: done

src/arweave/composite-client.ts
hlolli: done

djwhitt (approved):

PR #375, opened 2025-04-14
feat(arweave): add chunk broadcasting to weighted peer network [PE-7945]

hlolli:
- Introduced peer chunk broadcasting after secondary posting.
- Added support for weighted peer selection via `weightedPostChunkPeers`.
- Controlled via `ARWEAVE_PEER_CHUNK_POST_MIN_SUCCESS_COUNT_STRING` and `ARWEAVE_PEER_CHUNK_POST_CONCURRENCY_LIMIT_STRING`.
- Refactored `weightedChunkPeers` into `weightedGetChunkPeers` and `weightedPostChunkPeers` for clarity.
- Updated chunk fetching logic to use new `weightedGetChunkPeers`.
- Only HTTP 200 responses count as successful.

### DIFF
--- a/src/arweave/composite-client.ts
+++ b/src/arweave/composite-client.ts
@@ -84,7 +84,10 @@ interface Peer {
   lastSeen: number;
 }
 
-type WeightedPeerListName = 'weightedChainPeers' | 'weightedChunkPeers';
+type WeightedPeerListName =
+  | 'weightedChainPeers'
+  | 'weightedGetChunkPeers'
+  | 'weightedPostChunkPeers';
 
 export class ArweaveCompositeClient
   implements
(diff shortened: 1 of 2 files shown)


src/config.ts
```diff
@@ -174,6 +174,16 @@ export const CHUNK_POST_MIN_SUCCESS_COUNT = +env.varOrDefault(
   '3',
 );
 
+// Contribute to a healthy Arweave Network by posting chunks to peers
+// default value of 0 means this behaviour is disabled.
+export const ARWEAVE_PEER_CHUNK_POST_MIN_SUCCESS_COUNT_STRING =
```
djwhitt, line 179: The `_STRING` suffix on both environment variables should be removed. We only use that in cases where we are preserving the original string value rather converting it.

src/arweave/composite-client.ts
```diff
@@ -1253,6 +1258,88 @@ export class ArweaveCompositeClient
...
+                data: chunk,
+                timeout: DEFAULT_CHUNK_POST_RESPONSE_TIMEOUT_MS,
+                signal: AbortSignal.timeout(
+                  DEFAULT_CHUNK_POST_ABORT_TIMEOUT_MS,
+                ),
+                headers: originAndHopsHeaders,
+                validateStatus: (status) => status === 200,
+              });
+
+              if (resp.status === 200) {
```
djwhitt, line 1289: Is this redundant with `validateStatus` above?

hlolli: good catch, yes for sure it shouldn't be needed twice

src/config.ts
hlolli: fixed

src/arweave/composite-client.ts
hlolli: fixed

djwhitt (approved):

PR #395, opened 2025-05-22
feat(arns cache): preferred ArNS caching PE-8100

djwhitt:
Adds support for preferred caching based on user configurable lists of ArNS full names and base names (configured using PREFERRED_ARNS_NAMES and PREFERRED_ARNS_BASE_NAMES respectively). When preferred names are set and filesystem cache cleanup is enabled, the preferred names will be cleaned from the filesystem after
PREFERRED_ARNS_CONTIGUOUS_DATA_CACHE_CLEANUP_THRESHOLD instead of CONTIGUOUS_DATA_CACHE_CLEANUP_THRESHOLD.

This is accomplished by keeping an MRU list of ArNS names in the contiguous metadata cache and then checking it when cleaning cached data from the filesystem.

Headers for ArNS full, under, and base names were also added. In addition to being useful for end users. They are used to pass ArNS names through to the read through data cache so that they can be added to the metadata cache.


src/constants.ts
```diff
@@ -27,6 +27,9 @@ export const headerNames = {
   rootTransactionId: 'X-AR-IO-Root-Transaction-Id',
   dataItemDataOffset: 'X-AR-IO-Data-Item-Data-Offset',
   arnsTtlSeconds: 'X-ArNS-TTL-Seconds',
+  arnsName: 'X-ArNS-Name',
+  arnsBasename: 'X-ArNS-Basename',
+  arnsUndername: 'X-ArNS-Undername',
```
hlolli, line 32: unrelated to the PR, these are headers we want to add to exposed headers in cors config?

dtfiedler: cors dynamically supports all of the `values` in this constant https://github.com/ar-io/ar-io-node/blob/develop/src/app.ts#L58

src/data/read-through-data-cache.ts
```diff
@@ -65,6 +92,55 @@ export class ReadThroughDataCache implements ContiguousDataSource {
...
+    const existingMetadata = await this.metadataStore.get(hash);
+
+    // Prepare the metadata object for storage. Start with existing fields (if
+    // any) to preserve them.
+    const metadataToSet: ContiguousMetadata = {
+      ...(existingMetadata ?? {}),
+      accessTimestampMs: Date.now(), // Update the access timestamp
+    };
+
+    if (arnsName !== undefined) {
```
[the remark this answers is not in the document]
dtfiedler: we intentionally do not `await` to avoid blocking the event loop on cache writes - this is consistent with other cache writes throughout the codebase

src/data/read-through-data-cache.ts
```diff
@@ -157,8 +233,11 @@ export class ReadThroughDataCache implements ContiguousDataSource {
         (await this.contiguousDataIndex.getDataAttributes(id));
 
       if (attributes?.hash !== undefined) {
-        this.metadataStore.set(attributes.hash, {
-          accessTimestampMs: Date.now(),
+        const { arnsName, arnsBasename } = requestAttributes ?? {};
+        await this.updateMetadataCache({
```
dtfiedler, line 237: intended `await` here

src/middleware/arns.ts
```diff
@@ -123,7 +123,21 @@ export const createArnsMiddleware = ({
...
+        const parts = arnsSubdomain.split('_');
+        const basename = parts.pop(); // last part is basename
+        const undername = parts.join('_'); // everything else is undername
+        if (basename !== undefined && basename !== '') {
+          res.header(headerNames.arnsBasename, basename);
+        }
+        if (undername !== undefined && undername !== '') {
+          res.header(headerNames.arnsUndername, undername);
+        }
+      }
```
dtfiedler, line 139: nit: the resolver could return these values vs. computing them again here

src/data/read-through-data-cache.ts
djwhitt: @dtfiedler No, good call out. No need to `await` here.

src/middleware/arns.ts
djwhitt: Hmm... true. I'll check on patching this up.

PR #404, opened 2025-05-27
Switch to devenv

djwhitt:
## Summary
- add `devenv.nix` with the required packages
- update `flake.nix` to build the shell with `devenv`

## Testing
- `yarn test` *(fails: fetch failed)*


djwhitt:
Closing. We might revisit this in the future, but it was fiddly enough it's not worth it right now.

PR #413, opened 2025-05-29
feat(headers): add offset headers and sizes

dtfiedler:
These will be used for signature verification in wayfinder and are generally informative when requesting a data item data.


src/database/standalone-sqlite.ts
```diff
@@ -1107,6 +1114,12 @@ export class StandaloneSqliteDatabaseWorker {
       rootTransactionId,
       rootParentOffset: coreRow?.root_parent_offset,
       dataOffset: coreRow?.data_offset,
+      offset: dataItemAttributes?.offset,
+      itemSize: dataItemAttributes?.size,
+      signatureOffset: dataItemAttributes?.signature_offset,
+      signatureSize: dataItemAttributes?.signature_size,
+      ownerOffset: dataItemAttributes?.owner_offset,
+      ownerSize: dataItemAttributes?.owner_size,
```
dtfiedler, line 1123: @djwhitt - how do you feel about adding all these attributes on `ContiguousDataItemAttributes`?

src/types.d.ts
```diff
@@ -553,11 +553,17 @@ export interface ContiguousDataAttributes {
   hash?: string;
   dataRoot?: string;
   size: number;
+  offset: number;
   contentEncoding?: string;
   contentType?: string;
   rootTransactionId?: string;
   rootParentOffset?: number;
   dataOffset?: number;
+  itemSize?: number;
```
dtfiedler, line 563: this is intended to be the full size of the raw data item, naming could be better

src/constants.ts
```diff
@@ -26,6 +26,13 @@ export const headerNames = {
   cache: 'X-Cache',
   rootTransactionId: 'X-AR-IO-Root-Transaction-Id',
   dataItemDataOffset: 'X-AR-IO-Data-Item-Data-Offset',
+  dataItemSignatureOffset: 'X-AR-IO-Data-Item-Signature-Offset',
+  dataItemSignatureSize: 'X-AR-IO-Data-Item-Signature-Size',
+  dataItemOwnerOffset: 'X-AR-IO-Data-Item-Owner-Offset',
+  dataItemOwnerSize: 'X-AR-IO-Data-Item-Owner-Size',
+  dataItemRootParentOffset: 'X-AR-IO-Data-Item-Root-Parent-Offset',
+  dataItemOffset: 'X-AR-IO-Data-Item-Offset',
+  dataItemSize: 'X-AR-IO-Data-Item-Size',
```
dtfiedler: nit: in my testing,  we can return just data item offset and data item size to do verification via range requests for data items, but will need signature + owner offsets for non data items:

e.g. Data Item
```typescript
const response = await fetch(`http://localhost:4000/${txId}`, {
  method: 'HEAD',
});
const rootTransactionId = response.headers.get('x-ar-io-root-transaction-id');
const dataItemOffset = +response.headers.get('x-ar-io-data-item-offset');
const dataItemSize = +response.headers.get('x-ar-io-data-item-size');

// byte range for raw data request
const startByteRange = dataItemOffset;
const endByteRange = dataItemOffset + dataItemSize - 1;

const dataFromBundle = await fetch(
  `http://localhost:4000/${rootTransactionId}`,
  {
    method: 'GET',
    headers: {
      Range: `bytes=${startByteRange}-${endByteRange}`,
    },
  },
);

const maybeDataItem = new DataItem(dataFromBundle.data);
assert(await maybeDataItem.isValid(), 'Data item is not valid');
console.log('Data item is valid!')
```

dtfiedler: this way of verification via `DataItem` isn't ideal, though as it loads the full response into a Buffer, so the other additional offsets will be necessary for stream verification via `deepHash`

src/database/standalone-sqlite.ts
djwhitt: Yeah, we may want to do something different with the actual headers, but it's fine to return all of them here.

src/types.d.ts
djwhitt: We could call this `fullItemSize`.

src/constants.ts
djwhitt: For non data items we can fetch the Arweave TX headers from the chain.

PR #414, opened 2025-05-29
feat(verification): add prioritized data verification for preferred ArNS names

djwhitt:
This change introduces a prioritization system for data verification, focusing computational resources on preferred ArNS names while enabling flexible root transaction discovery.

Key changes:
- Add verification priority support to data content attributes, allowing items to be marked with priority levels
- Set verification priority to 80 for data accessed via preferred ArNS names (configured via PREFERRED_ARNS_NAMES and PREFERRED_ARNS_BASE_NAMES)
- Create DataItemRootTxIndex interface to abstract root transaction ID lookups
- Implement GraphqlRootTxIndex that queries trusted gateways via GraphQL, with support for nested bundles
- Add CompositeRootTxIndex that tries multiple index sources in sequence (local DB first, then GraphQL fallback)
- Enable background data verification by default with configurable MIN_DATA_VERIFICATION_PRIORITY (default: 80)
- Update verification queries to only process items meeting the minimum priority threshold

This approach ensures efficient use of resources by verifying only high-priority data while maintaining the ability to discover root transactions for items not yet indexed locally.


src/data/read-through-data-cache.ts
```diff
@@ -315,6 +316,16 @@ export class ReadThroughDataCache implements ContiguousDataSource {
 
               this.log.info('Successfully cached data', { id, hash });
               try {
+                // Check if the ArNS name is preferred
```
djwhitt, line 319: Factor this out into a method. Also, don't reference the config directly right here.

src/discovery/graphql-root-tx-index.ts
```diff
@@ -0,0 +1,193 @@
...
+
+    return currentId;
+  }
+
+  private async queryBundleId(
+    id: string,
+    log: winston.Logger,
+  ): Promise<string | null | undefined> {
+    // lower number = higher priority
+    const priorities = Array.from(this.trustedGateways.keys()).sort(
```
djwhitt, line 124: See if we can factor this out.

PR #422, opened 2025-06-05
feat(routes data): add digest and hash headers for data HEAD requests

djwhitt:
For GET requests we only return digest and etag headers when data is cached locally since we don't want to accidentally stream data from the network that doesn't match the headers. For HEAD requests this is not a concern since no actual data is returned.

- Add If-None-Match support for both HEAD and GET requests
- Fix ETag format to use proper quoted strings
- Extend digest/ETag headers to HEAD requests for previously

### DIFF
--- a/src/routes/data/handlers.ts
+++ b/src/routes/data/handlers.ts
@@ -41,11 +41,24 @@ const DEFAULT_CONTENT_TYPE = 'application/octet-stream';
 
 const REQUEST_METHOD_HEAD = 'HEAD';
 
+const handleIfNoneMatch = (req: Request, res: Response): boolean => {
+  const ifNoneMatch = req.get('if-none-match');
+  const etag = res.getHeader('etag');
+
+  if (ifNoneMatch !== undefined && etag !== undefined && ifNoneMatch === etag) {
+    res.status(304);
+    return true;
+  }
+  return false;
+};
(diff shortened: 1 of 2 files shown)


src/routes/data/handlers.ts
```diff
@@ -59,18 +72,27 @@ const setDigestStableVerifiedHeaders = ({
...
+    // request or cached locally. If the data is not cached, we might stream
+    // from the network and end up with a different hash than what is currently
+    // stored in the DB.
+    if (
+      dataAttributes.hash !== undefined &&
+      (data.cached || req.method === REQUEST_METHOD_HEAD)
+    ) {
       res.setHeader(headerNames.digest, dataAttributes.hash);
-      res.setHeader('ETag', dataAttributes.hash);
+      res.setHeader('ETag', `"${dataAttributes.hash}"`);
```
dtfiedler, line 84: why are the quotes necessary?

djwhitt: Claude pointed this out: https://stackoverflow.com/questions/66533278/why-is-etag-required-to-be-enclosed-in-double-quotes

djwhitt: Not _strictly_ necessary, but common practice.

PR #424, opened 2025-06-06
refactor(aws-client): require explicit AWS_REGION; drop us-east-1 default

hlolli:
* Remove silent 'us-east-1' region fallback.
* Initialise awsClient only when BOTH credentials and AWS_REGION are present;
  otherwise export `undefined`, preserving existing graceful-degrade path.
* Keeps optional-client semantics while preventing accidental cross-region calls.

### DIFF
--- a/src/aws-client.ts
+++ b/src/aws-client.ts
@@ -20,24 +20,26 @@ import awsLite from '@aws-lite/client';
 import awsLiteS3 from '@aws-lite/s3';
 import { fromNodeProviderChain } from '@aws-sdk/credential-providers';
 
-async function hasAnyAwsCreds() {
+async function canInitAwsClient() {
+  // 1. Region must be set
+  if (!process.env.AWS_REGION!) return false;
+
+  // 2. At least one credential source must resolve
   try {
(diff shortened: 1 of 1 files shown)


src/aws-client.ts
```diff
@@ -35,9 +35,16 @@ async function hasAnyAwsCreds() {
...
 export const awsClient = (await hasAnyAwsCreds())
   ? await awsLite({
-      // pass only the “static” settings; auth is left to the provider chain
       endpoint: process.env.AWS_ENDPOINT,
-      region: process.env.AWS_REGION, // optional – picked up from IMDS if omitted
+      // if you don't specify a region, it will default to us-east-1
+      // this matches the default region for aws-cli2
+      region:
+        process.env.AWS_REGION === undefined
+          ? 'us-east-1'
```
djwhitt: I would rather we require it to be specified than have a default. I think it's confusing if we have a default value in a situation (instance role) were region is usually picked up from metadata.

hlolli: ok I've chosen a different approach

PR #441, opened 2025-06-26
feat(data): add hash validation for peer data fetching PE-8238

djwhitt:
## Summary
- Added hash validation when fetching data from ar.io network peers
- Prevents accidental propagation of data that doesn't match local hashes
- Maintains backward compatibility with peers that don't support digest headers

## Changes
- Added `X-AR-IO-Expected-Digest` header constant
- Updated `ArIODataSource` to send expected hash in request headers when available
- Added validation in `parseResponse()` to check peer's `X-AR-IO-Digest` against expected hash
- Added comprehensive unit tests for all hash validation scenarios

## Test Plan
- [x] All existing tests pass
- [x] Added unit tests for hash validation:
  - Test successful data fetch when hashes match
  - Test rejection when hashes don't match
  - Test behavior when no expected hash is provided
  - Test backward compatibility when peer doesn't provide digest header
- [x] Build succeeds (`yarn build`)
- [x] Lint passes (`yarn lint:check`)

### DIFF
--- a/src/data/ar-io-data-source.ts
+++ b/src/data/ar-io-data-source.ts
@@ -15,6 +15,7 @@ import memoize from 'memoizee';
 
 import {
   ContiguousData,
+  ContiguousDataAttributes,
   ContiguousDataSource,
   Region,
   RequestAttributes,
@@ -319,11 +320,13 @@ export class ArIODataSource
 
   async getData({
     id,
+    dataAttributes,
     requestAttributes,
     region,
     retryCount,
   }: {
     id: string;
+    dataAttributes?: ContiguousDataAttributes;
     requestAttributes?: RequestAttributes;
     region?: Region;
     retryCount?: number;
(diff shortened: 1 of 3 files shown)


src/data/ar-io-data-source.ts
```diff
@@ -399,15 +408,31 @@ export class ArIODataSource
...
     requestStartTime: number;
     peer: string;
     ttfb: number;
+    expectedHash?: string;
   }): ContiguousData {
     const stream = response.data;
 
+    // Check if peer's digest matches our expected hash
+    if (expectedHash !== undefined) {
+      const peerDigest = response.headers[headerNames.digest.toLowerCase()];
```
hlolli, line 424: it says in axios docs https://axios-http.com/docs/res_schema

```
  // All header names are lower cased and can be accessed using the bracket notation.
```

but this isn't like computationally heavy thing to do, but this could be done without the toLowerCase strictly speaking.

hlolli (approved):

PR #453, opened 2025-07-25
Feature: add HyperBEAM sidecar

pierreneter:
allow the gateway to quickly run the HyperBEAM sidecar


djwhitt:
Hey, thanks taking the time make the HyperBEAM image and make this PR! We don't include 3rd party extensions in the ar-io-node due to the challenges of effectively managing the security of them on an ongoing basis, but we are happy to include this in the extension registry and I believe have already done so (https://github.com/ar-io/network-portal/issues/235).

PR #454, opened 2025-07-26
PE-8320: Update S3DataSource to support Turbo data items correctly

arielmelendez:
- Adds support for configuration of a secondary S3 bucket data source dedicated to vertical integration with Turbo
- Corrects the implementation for reading ranges and full items from s3 cache data that contain S3 metadata about the data item
- Adds support for zero-length data items stored in S3
- Adds tests for updated S3 features
- Removes automatic usage of the base S3 data source as a cache data store


src/data/s3-data-source.ts
```diff
@@ -166,3 +176,20 @@ export class S3DataSource implements ContiguousDataSource {
     }
   }
 }
+
+// Expected format: `bytes start-end/maxSize`
+function contentSizeFromContentRange(
+  contentRange: string | undefined,
+): number | undefined {
+  // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
```
djwhitt, line 184: Disabling because we want to concisely handle empty string and undefined?

package.json
```diff
@@ -10,7 +10,7 @@
   },
   "dependencies": {
     "@ar.io/sdk": "3.14.0",
-    "@aws-lite/client": "^0.22.4",
+    "@aws-lite/client": "^0.23.2",
```
djwhitt, line 14: `aws-lite` still good enough, eh?

src/system.ts
```diff
@@ -522,6 +535,8 @@ for (const sourceName of config.ON_DEMAND_RETRIEVAL_ORDER) {
   const dataSource = getDataSource(sourceName);
   if (dataSource !== undefined) {
     onDemandDataSources.push(dataSource);
+  } else {
+    log.error(`Data source ${sourceName} not found! Skipping.`);
```
djwhitt: Good catch. Might be worth throwing tbh.

djwhitt (commented):
Probably worth getting Claude to whip up some unit tests for this. I don't generally put a lot of stock in their initial coverage of functionality, but they still work pretty well for catching regressions.

src/data/s3-data-source.ts
arielmelendez: Yes, in this case.

package.json
arielmelendez: For now, it's sufficient. Might be a followup PR later on to move to full aws sdk for everything.

src/system.ts
arielmelendez: Yea it came up in testing. Probably should throw. Happy to update to that behavior.

arielmelendez: b211ef8

djwhitt (approved):

PR #476, opened 2025-08-21
feat: implement root parent and offset based data source with ANS-104 parsing (PE-8467)

djwhitt:
## Summary
Implements a data source that retrieves data items from ANS-104 bundles by combining root transaction lookup with on-demand bundle parsing to find data item offsets.

## Background
The gateway was unable to serve data items stored within ANS-104 bundles. When requesting a data item ID, the system would return a 404 error instead of finding the data item within its parent bundle.

## Implementation
This PR implements the architecture described in PE-8467:

### Components Added

1. **RootParentDataSource** (`src/data/root-parent-data-source.ts`)
   - Uses `DataItemRootTxIndex` to find the root transaction for any data item
   - Delegates to `Ans104OffsetSource` to parse bundles and find offsets
   - Wraps the underlying `ContiguousDataSource` to fetch data using the root ID and discovered offset
   - Passes through regular transactions unchanged

2. **Ans104OffsetSource** (`src/data/ans104-offset-source.ts`)
   - Parses ANS-104 bundle headers to locate data items
   - Supports recursive searching through nested bundles
   - Includes cycle detection to prevent infinite loops
   - Parses data item headers to extract only the data portion (excluding envelope)

### Key Features
- **Efficient parsing**: Only reads bundle headers, not full data items
- **Recursive support**: Handles arbitrary nesting of bundles
- **Correct data extraction**: Skips data item envelope (signature, owner, target, anchor, tags) to return only the actual data
- **Error handling**: Clear error messages and proper error propagation
- **Performance**: Header-only parsing minimizes data transfer

### Integration
The `TxChunksDataSource` is now wrapped with `RootParentDataSource` to enable transparent resolution of data items to their root transactions.

## Testing
- Added comprehensive unit tests for both components
- Tests cover: basic retrieval, nested bundles, cycle detection, edge cases
- All existing tests continue to pass

## Bug Fixes
During implementation, discovered and fixed an issue where data items were returning extra bytes (86176 instead of 85015) because the entire data item envelope was being included. The solution parses the data item header to calculate the correct offset to the actual data.

## JIRA Ticket
[PE-8467](https://ardrive.atlassian.net/browse/PE-8467)

[PE-8467]: https://ardrive.atlassian.net/browse/PE-8467?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/system.ts
+++ b/src/system.ts
@@ -16,8 +16,9 @@ import { GatewaysDataSource } from './data/gateways-data-source.js';
 import { ReadThroughDataCache } from './data/read-through-data-cache.js';
 import { SequentialDataSource } from './data/sequential-data-source.js';
 import { TxChunksDataSource } from './data/tx-chunks-data-source.js';
-import { RootParentDataSource } from './data/root-parent-data-source.js';
-import { Ans104OffsetSource } from './data/ans104-offset-source.js';
+// Temporarily disabled - will be re-enabled when other offset sources are available
+// import { RootParentDataSource } from './data/root-parent-data-source.js';
+// import { Ans104OffsetSource } from './data/ans104-offset-source.js';
 import { DataImporter } from './workers/data-importer.js';
 import { CompositeClickHouseDatabase } from './database/composite-clickhouse.js';
 import { StandaloneSqliteDatabase } from './database/standalone-sqlite.js';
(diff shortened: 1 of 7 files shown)


src/system.ts
```diff
@@ -428,12 +444,28 @@ export const chunkSource = new FullChunkSource(
...
 
+// Create ANS-104 offset source for parsing bundle headers
+const ans104OffsetSource = new Ans104OffsetSource({
+  log,
+  dataSource: baseTxChunksDataSource,
+});
+
+// Wrap the TX chunks data source with RootParentDataSource
+// This enables resolution of data items within ANS-104 bundles
+const txChunksDataSource: ContiguousDataSource = new RootParentDataSource({
```
djwhitt: Need to make this optional initially (we don't want to kill Goldsky with queries)

PR #485, opened 2025-09-11
feat: Add attributes-based parent traversal to RootParentDataSource

djwhitt:
## Summary

Implements efficient parent chain traversal using ContiguousDataAttributesStore with automatic fallback to legacy methods:

- **Attributes-first traversal**: Always attempts efficient traversal using cached attributes
- **Automatic offset accumulation**: Correctly calculates total offsets including dataOffset fields  
- **Leaf item size preservation**: Always requests correct data size from root bundles
- **Robust error handling**: Cycle detection, depth limits, and graceful fallback
- **Configurable fallback**: `ENABLE_LEGACY_ROOT_TRAVERSAL_FALLBACK` option (default: true)

### Data Flow
1. Request for data item → Try attributes-based traversal
2. Traverse parent chain → Accumulate offsets, remember original item size  
3. Reach root → Calculate final region with total offset + original size
4. Fetch from data source → Request exactly the leaf item size from root bundle
5. On failure → Optionally fall back to legacy traversal methods

### Performance Benefits
- Eliminates need to parse ANS-104 bundles for every data item request
- Uses cached attribute data for fast parent chain resolution
- Maintains backward compatibility through configurable fallback

## Test Plan

- [x] All existing tests pass (legacy functionality preserved)
- [x] New comprehensive test suite with 16 test cases covering:
  - [x] Single-level and multi-level parent traversals
  - [x] Self-referential roots and cycle detection
  - [x] Fallback behavior when attributes missing
  - [x] Error handling when fallback disabled
  - [x] Region handling with attributes traversal
  - [x] Proper size calculations and offset accumulation

## Configuration

New environment variable:
- `ENABLE_LEGACY_ROOT_TRAVERSAL_FALLBACK=true` (default) - Falls back to legacy methods when attributes incomplete
- `ENABLE_LEGACY_ROOT_TRAVERSAL_FALLBACK=false` - Throws error when attributes incomplete


src/data/composite-data-attributes-source.ts
```diff
@@ -0,0 +1,115 @@
...
+  async setDataAttributes(
+    id: string,
+    attributes: Partial<ContiguousDataAttributes>,
+  ): Promise<void> {
+    this.log.debug('Setting data attributes in cache', { id });
+    const existingAttributes = this.cache.get(id);
+    const mergedAttributes = existingAttributes
+      ? { ...existingAttributes, ...attributes }
+      : attributes;
+    // TODO: Clean up type assertion - cache should handle partial attributes properly
```
djwhitt: This looks like an outdated we can cleanup now.

src/data/turbo-dynamodb-data-source.ts
```diff
@@ -330,6 +368,31 @@ export class TurboDynamoDbDataSource implements ContiguousDataSource {
...
+        const attributes: Partial<ContiguousDataAttributes> = {
+          size: payloadLength,
+          dataOffset: payloadDataStart,
+          contentType: payloadContentType,
+          parentId: offsetsInfo.parentInfo.parentDataItemId,
+          offset: offsetsInfo.parentInfo.startOffsetInParentPayload,
+        };
+
+        // Set rootTransactionId if we have rootParentInfo
+        if (offsetsInfo.rootParentInfo !== undefined) {
```
djwhitt: I don't think this actually happens. We should remove this case.

src/data/ar-io-data-source.ts
```diff
@@ -328,6 +336,27 @@ export class ArIODataSource implements ContiguousDataSource {
       this.handlePeerSuccess(peer, kbps, ttfb);
     });
 
+    // Cache attributes discovered from peer headers
+    const peerDigest = response.headers[headerNames.digest.toLowerCase()];
+    if (peerDigest) {
```
djwhitt: We don't want to update attributes from peers (yet). We risk caching bad attributes.

PR #488, opened 2025-09-15
fix: preserve content type in RootParentDataSource for nested data items

djwhitt:
## Summary

Fixes an issue where the `RootParentDataSource` (offset-aware data source) was not preserving the content type of nested data items when resolving them from their root bundles. This caused responses to return the root bundle's content type instead of the actual data item's content type.

## Changes

- **Updated `traverseToRootUsingAttributes`** to track and return the original data item's `contentType` from its attributes
- **Modified both data fetching paths** to preserve the original content type:
  - **Attributes-based path**: Uses the content type discovered during traversal  
  - **Legacy fallback path**: Retrieves the original data item's attributes to get its content type
- **Added proper logging** for content type preservation in both paths
- **Used nullish coalescing** (`??`) to handle undefined content types gracefully

## Impact

This ensures that when serving nested data items from ANS-104 bundles, the response will have the correct content type of the actual data item rather than the content type of the root bundle. This is important for proper content handling by clients and browsers.

## Test Plan

- [ ] Verify that nested data items return their correct content type
- [ ] Test both attributes-based and legacy fallback paths
- [ ] Confirm that root bundles still return their own content types correctly
- [ ] Validate that fallback to root content type works when nested item has no content type

### DIFF
--- a/src/data/root-parent-data-source.ts
+++ b/src/data/root-parent-data-source.ts
@@ -66,9 +66,12 @@ export class RootParentDataSource implements ContiguousDataSource {
    * Traverses the parent chain using data attributes to find the root transaction.
    * Returns null if traversal is incomplete due to missing attributes.
    */
-  private async traverseToRootUsingAttributes(
-    dataItemId: string,
-  ): Promise<{ rootTxId: string; totalOffset: number; size: number } | null> {
+  private async traverseToRootUsingAttributes(dataItemId: string): Promise<{
+    rootTxId: string;
+    totalOffset: number;
+    size: number;
+    contentType?: string;
+  } | null> {
     const log = this.log.child({
       method: 'traverseToRootUsingAttributes',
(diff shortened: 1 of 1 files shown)


src/data/root-parent-data-source.ts
```diff
@@ -303,10 +319,16 @@ export class RootParentDataSource implements ContiguousDataSource {
...
+              originalContentType,
+              rootContentType: data.sourceContentType,
             },
           );
 
-          return data;
+          // Preserve the original data item's content type if available
+          return {
+            ...data,
+            sourceContentType: originalContentType ?? data.sourceContentType,
```
arielmelendez, line 330: Seems ok though it means if attributes failure fails intermittently we could end up using parent content types (which would likely be application/octet-stream, which is what I imagine we'd fall back to anyway?).

PR #545, opened 2025-11-17
chore(docker): remove fcon trusted checkpoint owner in docker-compose…

dtfiedler:
….ao.yaml


docker-compose.hb.yaml
```diff
@@ -9,6 +9,12 @@ services:
       - ${HB_CACHE:-./data/hyperbeam/cache-mainnet}:/opt/_build/genesis_wasm/rel/hb/cache-mainnet
     networks:
       - ar-io-network
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost:10000/~cron@1.0/every?interval=30-seconds&cron-path=qNvAoz0TgcH7DMg8BCVn8jF32QH5L6T29VjHxhHqqGE~process@1.0/now"]
```
arielmelendez, line 13: This should likely use the "once" cron rather than the every version.

docker-compose.ao.yaml
```diff
@@ -12,7 +12,7 @@ services:
       WALLET_FILE: ${CU_WALLET_FILE:-}
       ALLOW_PROCESSES: ${ALLOW_PROCESSES:-}
       ALLOW_OWNERS: ${ALLOW_OWNERS:-}
-      PROCESS_CHECKPOINT_TRUSTED_OWNERS: ${PROCESS_CHECKPOINT_TRUSTED_OWNERS:-fcoN_xJeisVsPXA-trzVAuIiqO3ydLQxM-L4XbrQKzY,-HFe6PleLxj1EdFMYMSetT2NIJioDsZIktn-Y0AwP54,WjnS-s03HWsDSdMnyTdzB1eHZB2QheUWP_FVRVYxkXk}
+      PROCESS_CHECKPOINT_TRUSTED_OWNERS: ${PROCESS_CHECKPOINT_TRUSTED_OWNERS:-WjnS-s03HWsDSdMnyTdzB1eHZB2QheUWP_FVRVYxkXk,-HFe6PleLxj1EdFMYMSetT2NIJioDsZIktn-Y0AwP54}
```
arielmelendez, line 15: I think we might also be able to include `3P3o-LPDuRV7pP5vyrx5ONrIMheZov0XVHj490z2fGk` which is the wallet Tom used to create that checkpoint for us just before we introduced the trie device.

PR #661, opened 2026-03-30
Feat/tx metadata and tag headers

vilenarios:
## Summary

  - **Tag Response Headers** — Data responses on `/raw/:id` and `/:id` now include `X-Arweave-Tag-*` HTTP headers with the transaction/data
   item's ANS-104 tags. Tags resolve in parallel with data retrieval with a configurable 5s timeout, so TTFB is unaffected. Opt-in via
  `ARWEAVE_TAG_RESPONSE_HEADERS_ENABLED=true`.
  - **`/tx/:id` for L2 Data Items** — `GET /tx/:id` now returns metadata for ANS-104 data items, not just L1 transactions. 3-tier lookup:
  local DB → on-demand binary header extraction (~8KB range fetch, not full bundle) → Arweave node fallback. Tiers 2 and 3 run in parallel
  with timeout.
  - **On-Demand Background Indexing** — When tag resolution times out or fails, a background resolve fires to index the item in the local
  GraphQL DB for future requests. Second request always has tags.
  - **Upstream Tag Propagation** — Captures `X-Arweave-Tag-*` headers from upstream ar.io gateways as a fallback when local resolution
  fails.
  - **ArNS Manifest Fix** — Fixed cache poisoning in `CompositeDataAttributesSource.setDataAttributes` where partial attributes from the
  stream overwrote the correct manifest content type from the DB, breaking ArNS manifest resolution.

  ## Key implementation details

  - Feature is opt-in: disabled by default, enable with `ARWEAVE_TAG_RESPONSE_HEADERS_ENABLED=true`
  - Tag sanitization prevents header injection (strips control chars, caps length at 128/4096)
  - `resolveTagsWithTimeout` properly cleans up timers on the fast path (no timer leak)
  - CORS simplified to `exposedHeaders: '*'` to cover all dynamic `X-Arweave-Tag-*` names
  - Tag resolution fires before `getDataAttributes` to maximize head start time
  - On-demand items persisted with `filter: '{"always": true}'` to satisfy NOT NULL constraint on `bundle_data_items.filter_id`
  - `DataItemMetaResolver` returns metadata for both L1 and L2 from local GraphQL DB
  - Envoy regex route sends only exact `/tx/{43-char-id}` to Node.js; sub-routes still go to Arweave nodes
  - `CompositeDataAttributesSource.setDataAttributes` no longer caches partials without an existing entry, preventing manifest cache
  poisoning
  - Env vars added to `docker-compose.yaml` and `.env.example`

  ## New environment variables

  | Variable | Default | Description |
  |----------|---------|-------------|
  | `ARWEAVE_TAG_RESPONSE_HEADERS_ENABLED` | `false` | Enable tag response headers (opt-in) |
  | `ARWEAVE_TAG_RESPONSE_HEADERS_MAX` | `100` | Max tag headers per response |
  | `ARWEAVE_TAG_RESPONSE_HEADERS_TIMEOUT_MS` | `5000` | Tag resolution + /tx/:id resolver timeout |

  ## Test plan

  - [x] 138 targeted tests passing (tag headers, upstream parsing, ArIO data source, composite attributes, arweave-tx, offset source)
  - [x] Full test suite: all failures are pre-existing on `develop` (SQLite infrastructure tests, DataVerificationWorker)
  - [x] Lint + TypeScript clean
  - [x] Live-tested on vilenarios.com:
    - L1 transactions: tags from LMDB txStore (sub-ms)
    - L2 data items (indexed): tags from local GraphQL DB (sub-ms)
    - L2 data items (un-indexed, small): tags resolve within 5s on cold load
    - L2 data items (un-indexed, large): tags on second request after background indexing
    - ArNS manifest resolution working (ardrive.vilenarios.com)
    - `/tx/:id` for both L1 and L2 with parallel Arweave fallback
    - On-demand indexed items visible in local GraphQL
    - Encrypted ArDrive files: Cipher/Cipher-IV tags present
    - CORS wildcard verified in browser DevTools

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -61,6 +61,8 @@ app.use(
     // Wildcard exposes all response headers to browsers (valid for non-credentialed
     // requests, which is all we have with Access-Control-Allow-Origin: *).
     // This covers ar-io headers, x402, tracing, and dynamic X-Arweave-Tag-* names.
+    // NOTE: If credentialed CORS is ever added, wildcard exposedHeaders is
+    // silently ignored by browsers per the spec — switch back to an explicit list.
     exposedHeaders: '*',
   }),
 );


src/data/ans104-offset-source.ts
```diff
@@ -889,4 +908,153 @@ export class Ans104OffsetSource {
...
+      if (tagsBytesLength > 0 && tagsCount > 0) {
+        bytes = await readBytes(reader, bytes, tagsBytesLength);
+        const tagsBytes = bytes.subarray(0, tagsBytesLength);
+        tags = deserializeTags(Buffer.from(tagsBytes));
+        const contentTypeTag = tags.find(
+          (tag) => tag.name.toLowerCase() === 'content-type',
+        );
+        contentType = contentTypeTag?.value;
+        headerOffset += tagsBytesLength;
+      }
```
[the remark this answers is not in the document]
vilenarios: Fixed.

src/data/composite-data-attributes-source.ts
```diff
@@ -106,9 +106,15 @@ export class CompositeDataAttributesSource
...
     this.log.debug('Setting data attributes in cache', { id });
     const existingAttributes = this.cache.get(id);
-    const mergedAttributes = existingAttributes
-      ? { ...existingAttributes, ...attributes }
-      : attributes;
-    this.cache.set(id, mergedAttributes as ContiguousDataAttributes);
+    if (existingAttributes != null) {
+      // Merge into existing complete entry — preserves DB-sourced fields
+      // like isManifest and contentType that partials may not include
+      this.cache.set(id, { ...existingAttributes, ...attributes });
```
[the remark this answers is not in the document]
vilenarios: Fixed.

src/lib/request-attributes.ts
```diff
@@ -88,6 +88,35 @@ export const generateRequestAttributes = (
...
+      const tagName = key.slice(TAG_HEADER_PREFIX.length);
+      // Headers may be string or string[] (multiple values with same name)
+      const values = Array.isArray(value) ? value : [value];
+      for (const v of values) {
+        tags.push({ name: tagName, value: v });
+      }
+    }
+  }
+
+  return tags.length > 0 ? tags : undefined;
```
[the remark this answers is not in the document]
vilenarios: Fixed.

docs/envs.md
```diff
@@ -285,3 +285,12 @@ These settings control chunk offset verification observations, which validate th
...
+
+| ENV_NAME                                         | TYPE                 | DEFAULT_VALUE                                 | DESCRIPTION                                                                                                                                                          |
+| ------------------------------------------------ | -------------------- | --------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| ARWEAVE_TAG_RESPONSE_HEADERS_ENABLED             | Boolean              | false                                         | If true, includes transaction/data item tags as `X-Arweave-Tag-*` and verification headers (`X-Arweave-Signature`, `X-Arweave-Owner`, etc.) on `/raw/:id` and `/:id` responses |
+| ARWEAVE_TAG_RESPONSE_HEADERS_MAX                 | Number               | 100                                           | Maximum number of tag headers to include per response. If a transaction has more tags, an `X-Arweave-Tags-Truncated: true` header is added                          |
```
djwhitt, line 296: This default limit is quite high. Have checked on the default header limits in things like nginx and CDNs?

envoy/envoy.template.yaml
```diff
@@ -256,6 +256,26 @@ static_resources:
                           route:
                             cluster: ario_gateways
                             timeout: 0s
+                        #! GET /tx/{id} routes to core for L1+L2 metadata.
```
djwhitt: I dislike overloading this since it alters the meaning of that path and creates a new commitment in the API. I'd rather see a transaction GraphQL resolver implemented for this use case.

src/data/data-item-meta-resolver.ts
```diff
@@ -0,0 +1,286 @@
...
+  target: string;
+  anchor: string;
+  tags: { name: string; value: string }[];
+  dataSize: number;
+  contentType?: string;
+  parentId?: string;
+  rootTransactionId?: string;
+  isStable?: boolean;
+  /** Internal: raw extraction data for persistence. Not used externally. */
+  _meta?: DataItemMeta & { itemOffset: number; itemSize: number };
```
djwhitt: We don't use the `_` convention anywhere else in the code. Just `meta` would be fine. That said, it's unclear to me what this is doing too. It looks like the contents of `meta` duplicates many of the fields in `ResolveDataItemMeta`. Why not include the offsets to the top level data structure and avoid the duplication?

src/data/composite-data-attributes-source.ts
```diff
@@ -100,15 +100,35 @@ export class CompositeDataAttributesSource
...
+      // that were set by the DB (contentType, isManifest) from being
+      // overwritten by partial producers (e.g., stream Content-Type).
+      const safeIncoming = { ...attributes };
+      if (existingAttributes.contentType != null) {
+        delete safeIncoming.contentType;
+      }
+      if (existingAttributes.isManifest === true) {
+        delete safeIncoming.isManifest;
+      }
+      this.cache.set(id, { ...existingAttributes, ...safeIncoming });
```
djwhitt: Could you avoid the above conditionals by reversing the order of the splats?

src/data/composite-data-attributes-source.ts
```diff
@@ -100,15 +100,35 @@ export class CompositeDataAttributesSource
     }
   }
 
+  /**
+   * Merge partial attributes into an existing cached entry. Authoritative
+   * fields (contentType, isManifest) are shielded from overwrite when already
+   * set. When no cached entry exists the partial is skipped so the next
```
djwhitt: > When no cached entry exists the partial is skipped so the next

This changes the behavior of the existing code. Is that intended?

src/routes/data/handlers.ts
```diff
@@ -315,18 +318,107 @@ export const shouldUsePrivateCacheControl = (
...
+ * Resolve metadata for a given ID from local stores only (fast path).
+ * Checks LMDB txStore and local GQL DB — both are sub-millisecond.
+ * Does NOT trigger remote resolution (root TX index, binary extraction).
+ */
+export const resolveItemHeaders = async (
+  id: string,
+  txStore: PartialJsonTransactionStore,
+  dataItemMetaResolver: DataItemMetaResolver,
+): Promise<ResolvedItemHeaders> => {
+  // Tier 1: L1 transaction from LMDB header store
```
djwhitt: LMDB is only used for local development. Production deployments typically cache headers in Redis. It's fine to try fetching from there, but that's an in-memory cache with a relatively short TTL. That said, the GQL DB query in the data item flow below will should return the metadata for L1 TXs too. It makes the naming of that a bit misleading though. It might be cleaner to combine all of this into a single call that handles both L1 and data items since GQL returns both.

src/data/data-item-meta-resolver.ts
```diff
@@ -0,0 +1,286 @@
...
+}
+
+export class DataItemMetaResolver {
```
djwhitt: The name of this is a little confusing. Is it a meta resolver (a resolver that resolves resolvers) or a meta [data] resolver? I know the answer, but it's not clear without more context.

src/lib/request-attributes.ts
```diff
@@ -88,6 +88,57 @@ export const generateRequestAttributes = (
...
+
+/**
+ * Parse X-Arweave-Tag-* headers from an upstream response into tag pairs.
+ * Accepts an optional rawHeaders array (alternating name/value pairs from
+ * Node.js http.IncomingMessage) to preserve duplicate headers that would
+ * otherwise be collapsed into comma-separated strings.
+ */
+export const parseUpstreamTagHeaders = (
+  headers: Record<string, string | string[]>,
+  rawHeaders?: string[],
```
djwhitt: I couldn't find anywhere `rawHeaders` was passed in. Is it being used?

docs/envs.md
djwhitt: Addressed — added `ARWEAVE_TAG_RESPONSE_HEADERS_MAX_BYTES = 8192` (nginx default) alongside a count limit of 100, and verification headers are prioritized over tags.

envoy/envoy.template.yaml
djwhitt: Addressed — the `/tx/:id` REST route was removed entirely (9150516e). Transaction metadata is now served via the GraphQL `transaction(id:)` resolver.

src/data/data-item-meta-resolver.ts
djwhitt: Addressed — the `_meta` nested object was flattened. Offset fields (`signatureOffset`, `ownerOffset`, `headerSize`, etc.) are now top-level optional fields on `ResolvedTxMetadata` (ea05296d).

src/data/composite-data-attributes-source.ts
djwhitt: Addressed — the fix builds an explicit `authoritative` object for `contentType` and `isManifest`, spread last: `{...existing, ...attributes, ...authoritative}`.

src/data/composite-data-attributes-source.ts
djwhitt: Addressed — the cache is now seeded with partial attributes when no entry exists, with a comment explaining the rationale (the next `getDataAttributes` call fetches the complete DB record and merges on top via `fetchAndCache`).

src/routes/data/handlers.ts
djwhitt: Addressed — resolution is now unified in `TxMetadataResolver` with a 3-tier fallback (LMDB txStore → LRU cache → GraphQL DB) that handles both L1 transactions and L2 data items in a single path.

src/data/data-item-meta-resolver.ts
djwhitt: Addressed — renamed to `TxMetadataResolver` in `tx-metadata-resolver.ts` (ea05296d), removing the ambiguity.

src/lib/request-attributes.ts
djwhitt: Confirmed unused — the `rawHeaders` parameter was removed entirely (0736a4a7).

src/data/data-item-meta-resolver.ts
```diff
@@ -0,0 +1,297 @@
...
+  target: string;
+  anchor: string;
+  tags: { name: string; value: string }[];
+  dataSize: number;
+  contentType?: string;
+  parentId?: string;
+  rootTransactionId?: string;
+  isStable?: boolean;
+  /** Internal: raw extraction data for persistence. Not used externally. */
+  _meta?: DataItemMeta & { itemOffset: number; itemSize: number };
```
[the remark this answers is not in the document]
djwhitt: Already addressed — the `_meta` nested object was flattened and offset fields are now top-level on `ResolvedTxMetadata` (ea05296d).

src/data/tx-metadata-resolver.ts
```diff
@@ -0,0 +1,297 @@
...
+        value: utf8ToB64Url(t.value),
+      })),
+      offset: m.itemOffset,
+      size: m.itemSize,
+      index: 0,
+      parent_id: rootTxId,
+      parent_index: 0,
+      root_tx_id: rootTxId,
+      root_parent_offset: 0,
+      filter: '{"always": true}',
```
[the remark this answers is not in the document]
djwhitt: This is fine — `{"always": true}` is an existing filter pattern used elsewhere in the codebase and no queries filter by the `filter` column in a way that would cause issues here.

src/routes/arweave-tx.ts
```diff
@@ -0,0 +1,119 @@
...
+    })),
+    target: meta.target,
+    quantity: '0',
+    data_size: String(meta.dataSize),
+    data_root: '',
+    reward: '0',
+    signature: meta.signature,
+    // Extra fields for data items
+    owner_address: meta.ownerAddress,
+    signature_type: meta.signatureType,
```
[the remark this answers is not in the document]
djwhitt: Addressed — `signatureType` is now surfaced through the GQL layer (added to schema, SQL queries, and row mappings), so `fromGqlTransaction()` passes through the value instead of hardcoding `undefined`.

src/data/data-item-meta-resolver.ts
```diff
@@ -0,0 +1,297 @@
...
+        return resolved;
+      }
+    } catch {
+      // DB lookup failed — not available locally
+    }
+
+    return undefined;
+  }
+
+  async resolve(id: string): Promise<ResolvedDataItemMeta | undefined> {
```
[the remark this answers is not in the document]
djwhitt: Addressed — added a `pendingPromises` map to coalesce concurrent cache misses, matching the existing pattern in `CompositeDataAttributesSource` (785bf107).

src/data/tx-metadata-resolver.ts
```diff
@@ -0,0 +1,297 @@
...
+    };
```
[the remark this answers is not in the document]
djwhitt: Addressed — `parent_id` now uses the last element of the `path` array (the immediate parent) for nested bundles, falling back to `rootTxId` only when `path` has a single element (0d3c42a0).

PR #692, opened 2026-04-28
fix(bundles): preserve optimistic data item back-fill (PE-9073)

vilenarios:
## Summary

Two interacting defects in `bundles.db` cause the indexer SQLite worker to crash and silently regress correctly back-filled state when the optimistic data-item admin endpoint (`/ar-io/admin/queue-data-item`, used by Turbo) is in use. Result on a live gateway: chain ingestion blocked at every flush-trigger height; 246k+ orphaned rows with NULL `parent_id`; 5.3 GB WAL.

### Defect A — flush violates NOT NULL on `stable_data_item_tags.parent_id`

`bundles/flush.sql:insertOrIgnoreStableDataItemTags` selected `new_data_items.parent_id` (nullable since the 2024-06-13 nullable-parent-and-root migration) into `stable_data_item_tags.parent_id` (NOT NULL, never relaxed). Whenever a tag row had a non-NULL `root_transaction_id` pointing at a now-stable block but the parent data item had `parent_id IS NULL`, `flushStableDataItems` raised `SQLITE_CONSTRAINT_NOTNULL`. The transaction rolled back, `saveBlockAndTxs` propagated the error, and block ingestion stopped advancing past that height.

### Defect B — `upsertNewDataItem` clobbers back-filled values with NULLs

`bundles/import.sql:upsertNewDataItem` ON CONFLICT DO UPDATE unconditionally overwrote `parent_id`, `root_transaction_id`, and `data_offset` with the incoming params. The admin endpoint at `routes/ar-io.ts` hard-coded `parent_id: null, root_tx_id: null, data_offset: null` in its spread, so every re-POST after an unbundle reset correctly back-filled values to NULL — exactly the input that fed Defect A on the next flush.

### Repro

1. POST a data item to `/ar-io/admin/queue-data-item` (NULL parent_id).
2. Wait for the indexer to unbundle the containing bundle (parent_id back-filled to bundle id).
3. POST the same data item again (parent_id reset to NULL by Defect B).
4. When the data item's tag row's `root_transaction_id` becomes part of a stable block, the next flush throws (Defect A); ingestion halts.

## Fix

- **`flush.sql`**: skip rows where `parent_id` or `root_transaction_id` IS NULL in both `insertOrIgnoreStableDataItems` and `insertOrIgnoreStableDataItemTags`. Orphans stay queryable in `new_*` until back-filled.
- **`import.sql`**: use `COALESCE(@param, existing)` for `parent_id`, `root_transaction_id`, and `data_offset` in `upsertNewDataItem` so NULL re-POSTs don't overwrite populated values. A block comment documents the remaining INSERT-only root-atom fields and the structural follow-up.
- **`standalone-sqlite.test.ts`**: regression test exercising the failure path (optimistic POST → unbundle back-fill → optimistic re-POST → flush), asserting all three columns survive AND the row lands in `stable_data_items` without throwing.

The `routes/ar-io.ts` widening that originally landed in commit `704039bf` was reverted in commit `6986343b` — admin endpoint stays a pure optimistic path with hard-coded NULLs for the bundling fields. Fix 2 alone is load-bearing for the chain-stall.

## Risk analysis

| Axis | Verdict |
|---|---|
| Correctness | Behavior changes only in the `@param IS NULL` upsert case (the bug) and the NULL-`parent_id` flush case (the crash). No other code path relied on either behavior. |
| Reorgs | `height-reset.sql` only nulls `height`; never touches `parent_id` / `root_transaction_id`. COALESCE doesn't interact with reorg handling. |
| Performance | Fix 1 adds index-friendly `IS NOT NULL` predicates; Fix 2 adds three `COALESCE` calls per upsert. Both negligible. |
| Schema | None. Statement-level changes only, no migration. |
| Caller compatibility | Verified the only callers of `saveDataItem` (`data-item-indexer`, `tx-metadata-resolver`, the admin route). None rely on clobber-to-NULL semantics. |

## Recovery

- **Orphans whose bundles haven't been unbundled yet**: heal automatically the next time their containing bundle is unbundled.
- **Orphans whose bundles were already unbundled and then NULL-clobbered by a later optimistic POST** (the 320/320 case in the original analysis): no automatic re-trigger. The new flush filter excludes them (good), but they linger in `new_data_items` until the height-threshold stale cleanup eventually deletes them. A one-shot repair (repopulate `parent_id` / `root_transaction_id` / `data_offset` from the corresponding `bundle_data_items` row) is filed as a follow-up.

## Out of scope (separate tickets)

- **Two-statement split for tuple consistency** (per review): the {parent_id, root_transaction_id, root_parent_offset, data_offset, offset, size, signature_offset, signature_size, owner_offset, owner_size, signature_type} fields form a "root atom" that should move atomically. Per-column COALESCE prevents NULL-clobber but doesn't enforce group consistency, and the remaining INSERT-only fields (root_parent_offset, the offset/size pairs, signature_type) are silently shadowed when admin POST creates the row before unbundle. Structural fix is to split `upsertNewDataItem` into `insertOptimisticDataItem` (admin path; never touches the tuple) and `upsertNewDataItemWithRootAtom` (unbundle path; atomic root-atom write with non-NULL `parent_id` validated). Out of scope for the chain-stall hotfix.
- **One-shot repair for already-clobbered rows** (the 320/320 case described above).
- **Tag-row PK** in `new_data_item_tags`: the current PK includes `root_transaction_id` so NULL accumulates distinct rows under repeated optimistic POSTs. Drop `root_transaction_id` from the PK; long tail behind WAL bloat.
- **`bundles.db-wal` size on the affected gateway** — should checkpoint normally once flushes stop crashing; monitor on rollout, manual `PRAGMA wal_checkpoint(TRUNCATE)` if needed.
- **Bundle queue backlog** (102k+ queued ∧ never unbundled on the affected host).

## Test plan

- [x] `yarn build`, `yarn lint:check`, `yarn test:ci` — green in CI.
- [x] Regression test exercises both Defect B (clobber preserved) and the headline crash from Defect A (flush completes, row lands in `stable_data_items`).
- [ ] Reviewer confirms no other code paths rely on clobber-to-NULL semantics in `upsertNewDataItem`.
- [ ] Deploy on the affected gateway (turbo-gw-fsn1-2): observe `flushStableDataItems` no longer throws; chain head advances past 1,905,193; WAL checkpoint resumes.

Refs: PE-9073

### DIFF
--- a/src/database/sql/bundles/flush.sql
+++ b/src/database/sql/bundles/flush.sql
@@ -21,6 +21,8 @@ FROM new_data_items ndi
 JOIN core.stable_block_transactions sbt
   ON ndi.root_transaction_id = sbt.transaction_id
 WHERE ndi.height < @end_height
+  AND ndi.parent_id           IS NOT NULL
+  AND ndi.root_transaction_id IS NOT NULL
 ON CONFLICT DO NOTHING;
 
 -- insertOrIgnoreStableDataItemTags
@@ -40,4 +42,6 @@ JOIN new_data_items ndi
 JOIN core.stable_block_transactions sbt
   ON ndit.root_transaction_id = sbt.transaction_id
 WHERE ndit.height < @end_height
+  AND ndi.parent_id           IS NOT NULL
+  AND ndi.root_transaction_id IS NOT NULL
 ON CONFLICT DO NOTHING;
--- a/src/database/sql/bundles/import.sql
+++ b/src/database/sql/bundles/import.sql
@@ -134,7 +134,7 @@ INSERT INTO new_data_items (
   @root_parent_offset
 ) ON CONFLICT DO
 UPDATE SET
-  height = IFNULL(@height, height),
-  root_transaction_id = @root_transaction_id,
-  parent_id = @parent_id,
-  data_offset = @data_offset
+  height              = IFNULL(@height,              height),
+  root_transaction_id = COALESCE(@root_transaction_id, root_transaction_id),
+  parent_id           = COALESCE(@parent_id,           parent_id),
+  data_offset         = COALESCE(@data_offset,         data_offset)
--- a/src/database/sql/bundles/import.sql
+++ b/src/database/sql/bundles/import.sql
@@ -120,6 +120,18 @@ UPDATE SET
   last_indexed_at = @indexed_at
 
 -- upsertNewDataItem
+--
+-- Tuple consistency note: the fields {parent_id, root_transaction_id,
+-- root_parent_offset, data_offset, offset, size, signature_offset,
+-- signature_size, owner_offset, owner_size, signature_type} describe a
+-- single bundling ("root atom") and ought to move atomically. The
+-- COALESCE clauses below cover only the three immediately involved in
+-- the PE-9073 chain-stall (parent_id, root_transaction_id, data_offset)
+-- plus the IFNULL on height. The remaining root-atom fields are
+-- INSERT-only — an admin POST that arrives before the unbundle path
+-- will leave them NULL on later unbundle writes. The structural fix
+-- (split this statement into an optimistic-only insert and a
+-- root-atom upsert) is tracked as a follow-up to PE-9073.
 INSERT INTO new_data_items (
   id, parent_id, root_transaction_id, height, signature, anchor,
   owner_address, target, data_offset, data_size, content_type,
--- a/src/database/standalone-sqlite.test.ts
+++ b/src/database/standalone-sqlite.test.ts
@@ -1599,11 +1599,47 @@ describe('StandaloneSqliteDatabase', () => {
         'data_offset must survive optimistic re-POST',
       );
       assert.deepEqual(row!.parent_id, fromB64Url(bundleParentId));
-      assert.deepEqual(
-        row!.root_transaction_id,
-        fromB64Url(bundleRootTxId),
-      );
+      assert.deepEqual(row!.root_transaction_id, fromB64Url(bundleRootTxId));
       assert.equal(row!.data_offset, 100);
+
+      // Now exercise the flush path that previously hit
+      // SQLITE_CONSTRAINT_NOTNULL on stable_data_item_tags.parent_id
+      // (Defect A). Force heights non-NULL on both the data item and
+      // its tag rows, then make bundleRootTxId stable so the flush JOINs
+      // match.
+      const dataItemHeight = 100;
+      bundlesDb
+        .prepare('UPDATE new_data_items SET height = @h WHERE id = @id')
+        .run({ h: dataItemHeight, id: fromB64Url(itemId) });
+      bundlesDb
+        .prepare(
+          'UPDATE new_data_item_tags SET height = @h WHERE data_item_id = @id',
+        )
+        .run({ h: dataItemHeight, id: fromB64Url(itemId) });
+      coreDb
+        .prepare(
+          `INSERT INTO stable_block_transactions (
+             block_indep_hash, transaction_id, block_transaction_index
+           ) VALUES (@hash, @tx, 0)`,
+        )
+        .run({
+          hash: crypto.randomBytes(32),
+          tx: fromB64Url(bundleRootTxId),
+        });
+
+      // Pre-fix this throws inside the worker transaction. With the fix
+      // (back-fill preserved by COALESCE; flush guarded by IS NOT NULL),
+      // it completes and the row lands in stable_data_items.
+      await db.flushStableDataItems(dataItemHeight + 1, Date.now());
+
+      const stableRow = bundlesDb
+        .prepare('SELECT id FROM stable_data_items WHERE id = @id')
+        .get({ id: fromB64Url(itemId) });
+      assert.notEqual(
+        stableRow,
+        undefined,
+        'data item should land in stable_data_items after flush',
+      );
     });
   });
 
--- a/src/database/standalone-sqlite.test.ts
+++ b/src/database/standalone-sqlite.test.ts
@@ -1506,6 +1506,107 @@ describe('StandaloneSqliteDatabase', () => {
     });
   });
 
+  describe('upsertNewDataItem clobber resistance (PE-9073)', () => {
+    // Regression: after the unbundle path back-fills parent_id /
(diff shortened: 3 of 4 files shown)


arielmelendez:
Thanks — flush filter and regression test are solid, and the chain-stall fix is what matters most right now. Two structural concerns before merge plus one nuance on the description.

## 1. Per-column `COALESCE` doesn't codify the consistency rule we need

The fields `{parent_id, root_transaction_id, root_parent_offset, data_offset, offset, size, signature_offset, signature_size, owner_offset, owner_size, signature_type, parent_index}` describe a single bundling — call it the **root atom**. They must be mutually consistent. A value from one bundling mixed with values from another is logically incoherent, and worse than missing information.

Per-column `COALESCE` only prevents NULL-clobber per field. It doesn't enforce that the group moves atomically, and it can't distinguish "caller has no claim about the bundling" from "caller is committing to a coherent bundling."

### The rule we want

Tuple states form a knowledge ladder; writers can only move up (or laterally between fully-known states for re-bundles):

- **Unknown** — all NULL. The optimistic resting state before any bundling claim has been made.
- **Fully-known** — root atom populated, `parent_id` non-NULL. `parent_id == root_transaction_id` means the data item is a direct child of the L1 bundle; `parent_id != root_transaction_id` means it lives in an intermediate nested bundle.

Allowed transitions:

- unknown → fully-known ✅
- fully-known → fully-known with different values (re-bundle: data item moves under a new parent) ✅
- fully-known → unknown ❌
- any write that NULLs out an existing populated tuple field ❌

Implication: `parent_id = NULL` is only ever a resting state — never a write target. A writer claiming a bundling commits to a non-NULL `parent_id` (equal to `root_transaction_id` if no nested parent).

### Bad workflows the current PR allows

**A. Stale partial admin POST corrupts a coherent unbundle.**

1. Unbundle writes `parent_id=Y, root_tx_id=Z, root_parent_offset=N` (coherent, from bundling V).
2. A caller POSTs admin with `parent_id=X` (stale, from an aborted bundling attempt) and `root_tx_id=null`.
3. Per-column `COALESCE` → `parent_id=X, root_tx_id=Z, root_parent_offset=N`. Mixed across two bundlings.

**B. Re-bundle interleaves with a stale optimistic POST.**

1. Data item is bundled into B1 → unbundle writes the B1 atom.
2. Same item is re-bundled into B2 → unbundle writes the B2 atom atomically.
3. An admin POST that arrived in flight with B1's `parent_id` and `root_tx_id=null` leaves `parent_id` from B1 alongside `root_tx_id` and offsets from B2.

**C. Pre-existing-but-related.** The `ON CONFLICT UPDATE` never touches `root_parent_offset`, the offset/size pairs, `signature_type`, or `parent_index`. So if admin POST creates the row first with NULLs for those, unbundle's later non-NULL values are silently dropped. Same family of bug; symptomatic of treating these as independent columns when they're really one tuple.

### Suggested shape

Split into two named statements rather than one upsert with conditional logic:

- **`insertOptimisticDataItem`** — used by the admin route. INSERT with the always-known optimistic columns (id, signature, owner, tags, target, anchor, indexed_at). On conflict, only updates the optimistic columns themselves — **never touches the tuple**. Handles "no bundling claim made."
- **`upsertNewDataItemWithRootAtom`** — used by the unbundle path (and any future caller that has full bundling knowledge). INSERT or full atomic overwrite of the root atom + a non-NULL `parent_id`. Handles "I know the bundling."

The "no NULL write over an existing value" property falls out for free: the optimistic statement never updates the tuple, so it can never NULL anything; the root-atom statement always writes a non-NULL `parent_id` by validation.

If a future caller eventually needs to commit to "I know my direct parent but not the bundling above it" (`parent_id`-only claim), it's a third statement that updates `parent_id` only when both existing `parent_id IS NULL` and `root_transaction_id IS NULL` — i.e., only as a transition out of unknown. Easy to add when that real caller exists; defer for now.

Asymmetry made explicit at the call site beats per-column flags or `CASE WHEN` gymnastics.

## 2. Please drop the route widening (or gate it behind §1)

The PR widens `QueueDataItemHeaders` to accept caller-provided `data_offset?`, `parent_id?`, `root_tx_id?` and forwards them. Today Turbo sends all-NULL so the bug is dormant — but once any caller starts sending partial values, the workflows in §1 above fire. The "defense in depth" framing actually weakens the invariant.

If there's no concrete current need for caller-provided values, drop this from the PR and keep the admin endpoint as a pure optimistic path. If a future caller does need it, gate it behind §1's two-statement split AND route-layer validation:

- A request providing any root-atom field must provide all of them, plus a non-NULL `parent_id`. Partial root-atom input → 400.
- A request providing `parent_id` only is rejected for now (defer until a concrete caller needs it).
- A request providing nothing about the bundling routes to `insertOptimisticDataItem`.

Without that validation, the route widening is the surface area for §1's bad workflows.

## 3. Healing claim needs a caveat

> Existing orphans heal automatically the next time their containing bundle is unbundled.

True for orphans whose bundles haven't been unbundled yet on this indexer. For orphans whose bundles were already unbundled and then clobbered NULL by a later Turbo POST (the 320/320 case in the original analysis), there's no automatic re-unbundle trigger. Those rows have non-NULL `height` (already protected by the existing `IFNULL`), so:

- The new flush filter excludes them — they don't migrate to stable.
- The 2-hour stale cleanup only deletes rows with `height IS NULL` or `height < threshold`.

So they linger until either Turbo POSTs them with non-null values (currently impossible without the route widening from §2), or the height becomes old enough to be cleaned up by the height-threshold path. Worth either filing a one-shot repair (repopulate from `bundle_data_items` where possible) or rewording the description to be explicit about the longer healing window.

## 4. Document the contract

Once the rule is settled, please add a short comment on the upsert SQL (or a section in `docs/`) naming the root atom and stating the knowledge-ladder transitions. Future contributors shouldn't have to re-derive this from a P1 incident.

## Tag PK — agreed out of scope, filing follow-up

+1 to deferring. Filing a separate ticket for: drop `root_transaction_id` from the `new_data_item_tags` PK and use `(tag_name_hash, tag_value_hash, data_item_id, data_item_tag_index)`. The current PK lets NULL accumulate distinct duplicate rows under repeated optimistic POSTs (`NULL ≠ NULL` in SQLite PKs); that's the long tail behind the WAL bloat beyond the immediate clobber.

PR #695, opened 2026-04-29
fix(arns): fall back to cached resolution on undefined fresh result (PE-9075)

vilenarios:
## Summary

- Fixes `CompositeArNSResolver` to fall back to a cached resolution when the fresh resolver returns `undefined` quickly, not only on `pTimeout` timeout.
- Matches the comment-documented intent ("fall back if error occurs OR timeout") that the previous implementation only partially honored.
- Adds `arns_cached_resolution_fallback_on_empty_total` counter for observability.

## Why

When AO/CU dry-runs fail fast (or the ArNS names cache fast-misses recently-registered names), `resolveParallel` resolves with `undefined` inside the timeout window. `pTimeout`'s `fallback` fires only on actual timeouts, so the gateway dropped through to `ARNS_NOT_FOUND_ARNS_NAME` and served the "Make this domain space yours" placeholder for names that had perfectly good (recently expired) cached resolutions.

Confirmed live on production: `sol.ar.io` and other names showing the placeholder during AO/CU flaps despite having usable cached resolutions in `KvArNSResolutionStore`.

## Test plan

- [x] `yarn test src/resolution/composite-arns-resolver.test.ts` — three new tests pass (9/9 in file)
- [x] `yarn test` (full suite) — 1703/1703 pass, 4 skipped, 0 fail
- [x] `yarn lint:check` — clean
- [x] `yarn build` — clean
- [ ] Manual smoke against staging gateway: trigger AO unreachability, observe cached fallback in span events (`Using cached resolution due to fresh-resolution returning no id`) and metric increments

## Risk / rollback

| Risk | Mitigation |
|---|---|
| Cached fallback returns very stale data when AO is down for hours | Bounded by `ARNS_CACHE_TTL_SECONDS` (default 30d). Operators can lower it for tighter freshness. |
| Cached fallback hides AO outages from monitoring | New `arns_cached_resolution_fallback_on_empty_total` counter makes it visible. Alert on rate-of-change. |
| Logs go quieter during AO flaps (old code emitted "Unable to resolve name" warning; new code only logs that when both fresh and cache are empty) | Use the new counter for alerting instead. |
| Test mocks don't catch a real edge case | Manual smoke against staging before merge. |

**Rollback:** revert the commit. The change is localized to one function in one file; revert is clean. No DB/state migration.

## Out of scope (deferred to separate PRs)

- **Bound max-staleness for cached fallback** (e.g., refuse cached if `resolvedAt + ttl * N < now`). Worth doing but adds a knob; address separately based on observed metric rates after merge.
- **Document `ARNS_CACHED_RESOLUTION_FALLBACK_TIMEOUT_MS` in `docs/envs.md`**. Should land but doesn't gate this fix.
- **Add a sibling counter for the existing timeout-fallback path** (currently bumps `usedCachedFallback` flag + span event but no Prometheus counter — observability is asymmetric across the two fallback triggers). 2-line follow-up.
- **Make `pTimeout` also handle errors as fallback triggers** (currently `resolveParallel` swallows errors to undefined; if it ever rethrows, the fallback would still need explicit handling). Left for now since the swallow-to-undefined behavior is consistent.

## Architectural notes for reviewers

- During sustained AO outages, the background-refresh path (`composite-arns-resolver.ts:319-337`) also returns `undefined` and never writes to the cache, so cached `resolvedAt` ages forever and every request hits the fallback. Pre-existing — not introduced here — but the new counter will make it visible: `arns_cached_resolution_fallback_on_empty_total` rising without corresponding cache-write activity is the signature.
- This fix also corrects a previously-broken case where `lastResolverTimeoutMs` (configurable, often >250ms) fires faster than `arnsCachedResolutionFallbackTimeoutMs`. Old code dropped to placeholder in that window; new code falls back to cached.

### DIFF
--- a/src/resolution/composite-arns-resolver.ts
+++ b/src/resolution/composite-arns-resolver.ts
@@ -348,8 +348,8 @@ export class CompositeArNSResolver implements NameResolver {
       // resolution returns no resolved id or we exceed the cached resolution
       // fallback timeout.
       let usedCachedFallback = false;
-      const fresh = await (cachedResolution
-        ? // Cached resultion exists
+      const resolution = await (cachedResolution
+        ? // Cached resolution exists
           pTimeout(
             this.resolveParallel({
               name,
@@ -379,20 +379,22 @@ export class CompositeArNSResolver implements NameResolver {
             parentSpan: span,
           }));
 
-      // pTimeout's `fallback` only fires on timeout — when it fires, `fresh`
-      // is the cached resolution returned by the fallback. Surface that path
-      // first so the on-empty metric below is never confused with a timeout.
-      if (usedCachedFallback) {
+      // pTimeout's `fallback` only fires on timeout — when it fires,
+      // `resolution` is the cached value returned by the fallback. Surface
+      // that path first so the on-empty metric below is never confused with
+      // a timeout. The `&& cachedResolution` guard narrows the type and is
+      // defensive against future drift around the ternary above.
+      if (usedCachedFallback && cachedResolution) {
         span.addEvent('Resolved by cache fallback');
-        return fresh as NameResolution;
+        return cachedResolution;
       }
       // If fresh resolution resolved fast with no resolved id (e.g.,
       // names-cache miss, AO/CU dry-run error swallowed to undefined), still
       // prefer the cached resolution if one exists. Matches the
       // comment-documented intent of "fall back if error occurs OR timeout".
-      if (fresh?.resolvedId !== undefined) {
+      if (resolution?.resolvedId !== undefined) {
         span.addEvent('Resolved by fresh resolution');
-        return fresh;
+        return resolution;
       }
       if (cachedResolution) {
         span.addEvent(
--- a/src/resolution/composite-arns-resolver.ts
+++ b/src/resolution/composite-arns-resolver.ts
@@ -344,10 +344,11 @@ export class CompositeArNSResolver implements NameResolver {
       metrics.arnsCacheMissCounter.inc();
       this.log.verbose('Resolution cache miss for ArNS name', { name });
 
-      // If there is a cached resolution, fall back to it if either an error
-      // occurs or we exceed the cached resolution fallback timeout
+      // If there is a cached resolution, fall back to it if either fresh
+      // resolution returns no resolved id or we exceed the cached resolution
+      // fallback timeout.
       let usedCachedFallback = false;
-      const resolution = await (cachedResolution
+      const fresh = await (cachedResolution
         ? // Cached resultion exists
           pTimeout(
             this.resolveParallel({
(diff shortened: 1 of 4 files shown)


src/resolution/composite-arns-resolver.ts
```diff
@@ -378,13 +379,27 @@ export class CompositeArNSResolver implements NameResolver {
...
-        } else {
-          span.addEvent('Resolved by fresh resolution');
-        }
-        return resolution;
+      // pTimeout's `fallback` only fires on timeout — when it fires, `fresh`
+      // is the cached resolution returned by the fallback. Surface that path
+      // first so the on-empty metric below is never confused with a timeout.
+      if (usedCachedFallback) {
+        span.addEvent('Resolved by cache fallback');
+        return fresh as NameResolution;
```
arielmelendez: I think the `as` here may subtly disguise situations where fresh is undefined. Might want to also assert in the conditional that fresh is defined in case any code around this changes.

arielmelendez: Adding to that, I think `fresh` as a name is confusing things here... I think we should keep it as `resolution`. It's a resolution that may or may not be fresh.

arielmelendez: Also why isn't the presence of `resolvedId` as important here as it is below? Here's how I'd simplify this whole thing:
```
if (resolved?.resolvedId !== undefined) {
  if (cachedResolution) {
    metrics.arnsCachedResolutionFallbackOnEmptyCounter.inc();
    span.addEvent('Resolved by cache fallback');
  } else {
        span.addEvent('Resolved by fresh resolution');
  }
     return resolved;  
} else {
  span.addEvent('Unable to resolve name');
  this.log.warn('Unable to resolve name against all resolvers', { name });
}
```
Something like that.

vilenarios: Thanks Ariel — addressed (1) and (2) in `22b21a95`:

- Renamed `fresh` → `resolution`. Agreed on the framing.
- Replaced `return fresh as NameResolution` with a runtime `usedCachedFallback && cachedResolution` narrowing. The cast was technically safe (pTimeout wrap only set up inside the `cachedResolution` ternary), but a runtime check is more defensive against future drift, which was your point.

On (3) — the simplification — I think the proposed shape has a subtle bug that drops the actual fix this PR is about, so I'd push back on adopting it verbatim. Two issues:

**A. False-positive on the `cache fallback` event/metric.** In the proposed shape, the inner `if (cachedResolution)` only checks whether *a* cached resolution exists in the variable, not whether we actually *returned* it. There's a real path where `cachedResolution` is defined but `resolution` is fresh-and-fine: cache exists but is expired (the early hot-cache return at L301–342 is bypassed because TTL elapsed), so the ternary takes the pTimeout branch, but `resolveParallel` completes successfully *within* `arnsCachedResolutionFallbackTimeoutMs`. Then `resolution` is fresh, `cachedResolution` is also defined, and the proposed code would emit `'Resolved by cache fallback'` and bump `arns_cached_resolution_fallback_on_empty_total` — even though we served fresh data.

**B. The proposed `else` branch loses the bug fix.** When `resolution?.resolvedId === undefined` AND `cachedResolution` is defined, this PR's whole purpose is to *return cachedResolution* (with the on-empty counter bump). The proposed `else` only logs and falls through, leaving the gateway dropping to `ARNS_NOT_FOUND_ARNS_NAME` — same bug as before.

The three behaviors the function needs to keep distinct are:

| Condition | Return | Span event | Metric |
|---|---|---|---|
| pTimeout fallback fired (timeout) | `cachedResolution` | Resolved by cache fallback | — |
| `resolution.resolvedId` defined (fresh ok) | `resolution` | Resolved by fresh resolution | — |
| `resolution.resolvedId` undefined AND cache exists | `cachedResolution` | Using cached resolution due to fresh-resolution returning no id | `arns_cached_resolution_fallback_on_empty_total++` |
| Neither | empty `NameResolution` (existing fallthrough) | Unable to resolve name | — |

Distinguishing the first row from the second requires either tracking `usedCachedFallback` or comparing `resolution === cachedResolution` by identity (equivalent under pTimeout's contract). That's why I kept the explicit branches. Open to other shapes if you see one I'm missing — happy to keep iterating.

arielmelendez (approved):

PR #702, opened 2026-05-04
feat(httpsig): bind body to signed response via Content-Digest for small uncached + chunks

vilenarios:
## Summary

Today the data response path emits `Content-Digest` only for cached responses and HEAD requests — see the comment at `handlers.ts:195-198`. Uncached `/raw/<id>` GETs go out signed with no body-binding, so a client can verify the signature but not that the body wasn't tampered with at the gateway. This PR closes that gap for small bodies and for chunks.

`Content-Digest` is already in `CO_SIGNABLE_HEADERS` (`src/lib/httpsig.ts:56`), so emitting it more often automatically extends the HTTPSIG signature's reach to include the body — no signature-shape change.

## Changes

- **New helper** `src/routes/data/buffered-digest.ts`. When an uncached response declares `data.size <= HTTPSIG_BODY_DIGEST_BUFFER_MAX_BYTES` (default 2 MiB), buffer the stream while updating SHA-256, then emit `Content-Digest` + `ETag` + `Content-Length` and write the buffered bytes. Larger or unknown-size bodies stream as today.
- **3 full-body GET sites** in `src/routes/data/handlers.ts` swap `pipeStreamToResponse` → `sendBodyWithOptionalDigest`. Range path untouched (partial bodies shouldn't carry full-body digest).
- **Chunk path** (`src/routes/chunk/handlers.ts`) extended: chunks are bounded at 256 KiB and already in memory, so SHA-256 inline (~50 µs) whenever cache didn't pre-populate `hashString`. Every served chunk now carries a body-bound digest.
- **New env** `HTTPSIG_BODY_DIGEST_BUFFER_MAX_BYTES` (default `2097152` = 2 MiB; set `0` to disable the buffered branch).
- **New metric** `httpsig_content_digest_total{source}` with 6 labels (`cache_hit`, `computed_buffered`, `skipped_size_unknown`, `skipped_too_large`, `skipped_disabled`, `overran_threshold`) so the threshold can be tuned from real traffic.

## Why 2 MiB default

Coverage vs TTFB tradeoff. At 1 MiB you skip most modern JS bundles. At 5 MiB you start adding meaningful TTFB tax (~500 ms on the unlucky uncached miss for a 10 MB/s upstream). 2 MiB catches typical webpages, JS bundles, and most images while keeping per-request memory bounded (100 concurrent × 2 MiB = 200 MB peak vs 8 GB Node heap). Operators can tune from the metric — bump if `skipped_too_large` is dominated by JS bundles (verification matters), keep at 2 MiB if dominated by media (less useful as a security primitive).

## Test plan

- [x] **8 unit tests** in `src/routes/data/buffered-digest.test.ts` — golden vector, all skip branches, size-lie overrun fallback, exact-boundary, cache-hit short-circuit, multi-size property check.
- [x] **3 contract-change updates** in `handlers.test.ts` — old assertions said "no digest for uncached"; that contract is what's changing. New assertions verify computed digest matches served body.
- [x] Type check clean for changed files.
- [x] Existing `httpsig.test.ts`, `digest.test.ts`, `handlers.test.ts` (other tests) regression-pass.
- [ ] Live perf test against local build using the bench harness — coming next.

## Operator notes

- **Default on** at 2 MiB.
- **Rollback**: set `HTTPSIG_BODY_DIGEST_BUFFER_MAX_BYTES=0` in `.env`, force-recreate `core`. No code change needed.
- **Monitoring**: watch `httpsig_content_digest_total{source}` to verify branches fire as expected; `httpsig_signing_duration_seconds` should stay in the ~70-100 µs range (signing cost itself unchanged — buffering is a separate path).

## Out of scope (deliberate)

- Don't sign body bytes directly with Ed25519 — `Content-Digest` is the right primitive, this PR just emits it more often.
- Don't try streaming-hash with delayed headers — TTFB tax for large bodies is real, skip path is correct.
- Don't change the signed-headers list — `content-digest` was already in `CO_SIGNABLE_HEADERS`.


vilenarios:
Closing — this work already shipped on develop via #699.

  The buffered-digest helper and supporting wiring (`HTTPSIG_BODY_DIGEST_BUFFER_MAX_BYTES`,
  `httpSigContentDigestTotal`, `httpSigBufferedBytesInflight`, the 3 call sites in
  `src/routes/data/handlers.ts`, the chunk-handler digest emission) were absorbed
  into the #699 branch and merged at 1a6b7c46. From there the implementation was
  polished further:

  - `5a54cbf8` — switched 502-header stripping from denylist → allowlist (future-proof
    against new trigger headers leaking onto error responses); added typed
    `incHttpSigContentDigest(...)` wrapper with a `HttpSigDigestSource` union.
  - `2ec59c94` — added the `short_read` 502 path (catches the silent-truncation
    case where a source declares N bytes but ends after K<N); tightened
    `HTTPSIG_BODY_DIGEST_BUFFER_MAX_BYTES` validation to `^\d+$` +
    `Number.isSafeInteger`; restored `createTestLogger` per CLAUDE.md; plus TSDoc
    on the new exports.

  `git diff origin/develop feat/httpsig-content-digest-buffered` for
  `src/routes/data/handlers.ts` and `src/routes/data/handlers.test.ts` is 0 lines.
  The remaining conflicts only exist because this branch holds older drafts of
  buffered-digest.ts, chunk handlers, config validation, and metrics — taking any
  of them would regress what #699 has since improved.

  Branch can be deleted.

PR #719, opened 2026-05-12
perf(bundles): index retry_attempt_count for selectFailedBundleIds

vilenarios:
## Summary

- Replace `import_attempt_last_retried_idx` with `bundles_active_retry_priority_idx` on `(last_fully_indexed_at, retry_attempt_count, last_retried_at)` — the column the retry-loop query actually orders by
- The pre-existing index has been on the wrong column (`import_attempt_count`) since the Jan 2025 retry-stats refactor (commit 528e1840 / migration `2025.01.30T14.12.03.bundles.add-retry-stats.sql`) — the `ORDER BY` clause was updated to use `retry_attempt_count` but the index was not, so the planner has been falling back to a temp B-tree on millions of rows for the last ~15 months
- `import_attempt_count` is still maintained (incremented in `import.sql`) and exposed via `/ar-io/admin/bundle-status/:id`, but that lookup is served by the primary-key index, so dropping the old retry-loop index is safe

## Why a non-partial composite

A smaller partial form `(retry_attempt_count, last_retried_at) WHERE last_fully_indexed_at IS NULL` was tested first but the SQLite planner ignores it for this query. The `WHERE` clause in `selectFailedBundleIds` contains an `OR`, which causes the planner to prefer MULTI-INDEX OR + temp sort over scanning a partial index. The non-partial composite with `last_fully_indexed_at` as the leading column lets the planner equality-search on `IS NULL` and walk the remaining entries in `(retry_attempt_count, last_retried_at)` order, eliminating the temp B-tree for the inner `ORDER BY`.

## EXPLAIN QUERY PLAN before/after

**Before:**
```
SEARCH TABLE bundles USING INDEX bundles_last_fully_indexed_at_idx (last_fully_indexed_at=?)
USE TEMP B-TREE FOR ORDER BY      ← building a B-tree on ~10M matched rows every cycle
```

**After:**
```
SEARCH TABLE bundles USING INDEX bundles_active_retry_priority_idx (last_fully_indexed_at=?)
(no inner temp B-tree)
USE TEMP B-TREE FOR ORDER BY      ← outer ORDER BY RANDOM() over 5,000 rows; cheap and unavoidable
```

## Measured impact

| Environment | Before | After | Speedup |
|---|---|---|---|
| Offline (cp of ~10M-row bundles.db, cold cache, ANALYZE'd) | 1,735 ms | 5 ms | ~347× |
| Live (under concurrent writes/WAL contention) | 4,320 ms | 20.5 ms | ~211× |

Bonus observed: downstream queries that share the SQLite worker (`updateBundlesFullyIndexedAt`, `updateBundlesForFilterChange`) also got 2-8× faster after the change — likely from reduced WAL contention now that the retry loop's pick is effectively free.

## Storage impact

- Drop `import_attempt_last_retried_idx`: -133 MB
- Add `bundles_active_retry_priority_idx`: +128 MB
- Net change: ≈ -5 MB of index space, no file-size growth (SQLite reuses freed pages)

The new index ends up slightly smaller because the leading `last_fully_indexed_at` column is mostly `NULL`, which SQLite encodes in 1 byte.

## Migration runtime

On a 9.8M-row `bundles.db` (during a clean core stop/start cycle):
- `DROP INDEX`: ~2.8 s
- `CREATE INDEX`: ~7.4 s
- `ANALYZE bundles`: ~12.4 s
- Total: ~25 s

Operators with significantly larger DBs should expect proportionally longer migration time. The migration runs once at gateway startup (umzug) and blocks core startup until complete.

## Test plan

- [x] Migration SQL is idempotent (`DROP IF EXISTS` / `CREATE IF NOT EXISTS`) — verified by running the up migration twice against a fresh DB
- [x] Down migration restores the prior index — verified by running up → down → verifying schema matches the pre-migration `test/bundles-schema.sql`
- [x] EXPLAIN QUERY PLAN confirms the new index is picked up automatically after ANALYZE — no `INDEXED BY` hint needed
- [x] Live-deployed on a 9.8M-row gateway; 25-min post-restart observation shows expected speedup with no errors or regressions
- [x] `test/bundles-schema.sql` updated to match the new index definition
- [ ] `yarn test` (could not run in worktree due to local toolchain — needs verification by reviewer)
- [ ] Confirm no regression in other repair-worker queries (`updateFullyIndexedAt`, `updateForFilterChange`) on a clean reviewer environment

## Notes for reviewers

- The two commits can be squashed on merge
- This is a low-risk change (single SQL migration, fully reversible via the down migration). Should be safe to ship in any release without coordination
- Down migration re-creates the prior (broken) index as-is, for true rollback symmetry. If a reviewer prefers the down to instead leave the system index-less, happy to adjust

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- The bundle retry-loop SELECT (`selectFailedBundleIds`) is now ~200× faster
+  after correcting the index column. The pre-existing
+  `import_attempt_last_retried_idx` had been on `import_attempt_count` since
+  the Jan 2025 retry-stats refactor, but the query orders by
+  `retry_attempt_count` — a different column. Replaced with
+  `bundles_active_retry_priority_idx` on `(last_fully_indexed_at,
+  retry_attempt_count, last_retried_at)`. Live measurement: 4.32 s → 20.5 ms
+  per call.
+
 ## [Release 78] - 2026-05-08
 
 This is a **recommended release** focused on **request-cancellation
--- a/migrations/2026.05.12T02.00.00.bundles.fix-retry-attempt-index.sql
+++ b/migrations/2026.05.12T02.00.00.bundles.fix-retry-attempt-index.sql
@@ -1,0 +1,39 @@
+-- up migration
+--
+-- Replace `import_attempt_last_retried_idx` with an index that supports the
+-- actual `selectFailedBundleIds` query in `src/database/sql/bundles/repair.sql`.
+--
+-- Root cause: the Jan 2025 retry-stats refactor (commit 528e1840,
+-- migration 2025.01.30T14.12.03.bundles.add-retry-stats.sql) changed
+-- the query's `ORDER BY` from `import_attempt_count, last_queued_at` to
+-- `retry_attempt_count, last_retried_at` and updated the trailing column
+-- of the index from `last_queued_at` to `last_retried_at`, but left the
+-- leading column as `import_attempt_count` instead of swapping it to
+-- `retry_attempt_count`. The Feb 2025 follow-up (2025.02.12T21.11.24)
+-- only added the partial-index predicate; it perpetuated the mismatch.
+--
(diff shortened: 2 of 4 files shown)


vilenarios:
@coderabbitai review

PR #720, opened 2026-05-12
perf(data): fix streaming backpressure + add keep-alive HTTP agents

vilenarios:
## Summary

Two related fixes targeting external-memory bloat in the bundle data fetch path. At high `ANS104_DOWNLOAD_WORKERS` (24+) the gateway's container RSS climbed from ~2 GB to 14-18 GB over 14-25 min, after which the cgroup OOM-killer terminated the process. Sampling during a live incident captured the exact growth curve; both root causes traced to the data-source layer.

### 1. `ReadThroughDataCache`: fix streaming backpressure on cache miss

On cache miss, the code piped `data.stream` into the disk-cache write stream via `pipeline()`, then ALSO attached a `.on('data')` listener for hashing and byte counting. Two consumers on the same readable forced it into flowing mode and short-circuited the pipeline's backpressure — `cacheStream`'s internal buffer grew well beyond `highWaterMark` while waiting on disk writes, holding multi-MB of pending data per concurrent download.

`setMaxListeners(Infinity)` was already in place, silencing the very warning intended to flag this pattern.

Fix: move hashing + counting into a `Transform` stream and put it inline in the pipeline (`data.stream → hashingStream → cacheStream`) so backpressure flows end-to-end through the standard writable mechanism.

### 2. `GatewaysDataSource`: shared keep-alive HTTP agents

`axios.create()` was called per request with no agent configuration, so every request used Node's default agent with `keepAlive: false`. Each request opened a fresh TCP+TLS connection that closed immediately after the response, sending sockets through TIME_WAIT (~60s kernel state per tuple). Live `ss -s` showed 500+ closed sockets accumulated under load.

Fix: per-gateway-URL agent cache. Each gateway URL gets one long-lived `http.Agent` or `https.Agent` with `keepAlive: true, maxSockets: 16, maxFreeSockets: 4`. Connections are reused across requests, TLS sessions can be resumed, TIME_WAIT pressure drops sharply.

Note on `instanceof`: `https.Agent extends http.Agent`, so an `instanceof http.Agent` check would treat both as http. Used `gatewayUrl.startsWith('https://')` instead.

## Why these are independent but related

The streaming backpressure bug is the dominant source of in-flight Buffer accumulation per worker. The agent change reduces handshake cost and kernel state churn — modest standalone improvement, but at high worker counts it compounds with the streaming fix to keep memory stable.

## Empirical results

Deployed locally on a production-scale gateway (9.8M-bundle DB, 24 download workers, ArDrive filter, kernel sysctl tuned, log rotation in place). Compared to pre-fix behavior at the same worker count.

| Metric | Pre-fix at 24 workers | Post-fix at 24 workers |
|---|---|---|
| Container restart cadence | every **14-25 min** | **2h50min+ uptime** observed, no kernel-OOM crashes |
| Container RSS growth rate | ~480 MB/min (catastrophic) | ~22 MB/min (slow, manageable) |
| Container RSS at point of failure | 14-18 GB → cgroup OOM | n/a — within container budget |
| `process_resident_memory_bytes` after 90 min | OOM'd long before | 6.0 GB stable-ish, slow growth |
| V8 `nodejs_heap_size_used_bytes` after 90 min | n/a | 2.76 GB |
| `uncaught_exceptions_total` per hour | 6-15 (mostly `ENOBUFS`) | < 1 |
| `ss -s` TIME_WAIT sockets | 185+ | 132 (-28 %) |
| `ss -s` closed sockets | 509 | 495 |
| Sustained bundle unbundle rate | not sustainable (kept crashing) | ~1,400-2,000 bundles/h |

The streaming-backpressure fix eliminated the dominant external-memory leak; the agent change reduced TCP/TIME_WAIT churn proportionally. The two combine to take the gateway from a hard 14-25 min crash cycle to multi-hour productive lifecycles.

## Residual leak (out of scope for this PR — follow-up)

After the fixes, V8 heap growth is now the dominant memory pressure (was external Buffer growth before). Heap climbs ~22 MB/min in a long-running lifecycle. Linear extrapolation to the `--max-old-space-size=16384` ceiling: ~12 h between Node-level OOM exits. Suspected cause is JS object retention in caches or queues (filter-result cache, ArNS resolution cache, or per-bundle metric maps without eviction); needs separate investigation. Mentioning here for completeness — does not block this PR, which solves a much faster failure mode.

## Test plan

- [x] Code-reading verification — confirmed dual-consumer pattern in `ReadThroughDataCache`; confirmed no agent config in `GatewaysDataSource`
- [x] `axios.create()` agent assignment uses protocol prefix check (not `instanceof http.Agent`)
- [x] `pipeline()` chain (data.stream → hashingStream → cacheStream) preserves error semantics — single error handler in the pipeline callback
- [x] Local deployment on a production-scale gateway; 2h50min+ uptime at 24 workers vs 14-25 min pre-fix
- [x] Memory profile recorded (table above) — external-memory leak eliminated, V8 heap growth slow but present (residual leak noted as follow-up)
- [x] CI: `yarn test` passing on both ubuntu-latest and macos-latest (re-run after prettier formatting commit)
- [x] CI: prettier-check passing (third commit normalizes formatting around the multi-line pipeline call)

## Commit structure

- `4c9d1d13` `perf(data): fix streaming backpressure in ReadThroughDataCache cache-miss path`
- `f487dcaa` `perf(data): add shared keep-alive HTTP agents to GatewaysDataSource`
- `ee10e120` `style: apply prettier formatting`

Easy to squash on merge, or to revert individually if needed.

## Stacks on

Based on `perf/bundles-retry-attempt-index` (PR #719). Once that merges, GitHub will auto-rebase this onto develop.


vilenarios:
@coderabbitai review

PR #726, opened 2026-05-13
perf(bundles): parallel retry lane for large Turbo bundles

vilenarios:
## Summary

Adds a second SQL statement + retry loop in \`BundleRepairWorker\` that targets bundles with **known** \`data_item_count >= threshold\` (default 50 — above direct ArDrive's 1-2 DIs, captures Turbo bulk). It runs alongside the existing retry loop on its own setInterval.

**Why**: the existing retry loop selects bundles ordered by \`retry_attempt_count\` — treating a 2-DI direct ArDrive bundle and a 500-DI Turbo bundle equally. On a typical ArDrive gateway, small direct bundles outnumber Turbo bundles ~50:1 (recent 24h on a test gateway: 11,454 small unbundled, 1 large). The small ones dominate the worker pool and large Turbo bundles starve, even when known and in flight. DI throughput suffers as a result — most of the ArDrive data volume lives in those 100-1000-DI Turbo bundles.

**Why not just change the existing ORDER BY** to \`data_item_count DESC\`? Three independent reasons:
1. **Starvation**: brand-new small uploads would wait days behind the Turbo backlog (~92k known small bundles still in flight at change time).
2. **NULL handling**: 2.5M matched bundles haven't been downloaded yet, so \`data_item_count IS NULL\`. \`NULLS LAST\` puts the bulk of pending work at the back of the queue.
3. **Regresses PR #719**: the existing query's index (\`bundles_active_retry_priority_idx\` on \`(last_fully_indexed_at, retry_attempt_count, last_retried_at)\`) wouldn't satisfy the new sort order, reintroducing the TEMP B-TREE FOR ORDER BY that PR #719 fixed (211× live measurement).

## Design

- **New SQL**: \`selectLargeFailedBundleIds\` (added to \`src/database/sql/bundles/repair.sql\`). Same outer WHERE as \`selectFailedBundleIds\` plus \`data_item_count IS NOT NULL AND data_item_count >= @threshold\`, ORDER BY \`data_item_count DESC, retry_attempt_count, last_retried_at ASC\`.
- **New partial index**: \`bundles_large_active_priority_idx ON bundles (data_item_count, retry_attempt_count, last_retried_at) WHERE last_fully_indexed_at IS NULL AND data_item_count >= 50\`. Tiny — partial WHERE keeps it to ~1k rows vs the 9.8M-row table.
- **New retry method**: \`BundleRepairWorker.retryLargeBundles\` — same shape as the existing \`retryBundles\`, just calls the new SQL.
- **Three new env vars**: \`BUNDLE_REPAIR_LARGE_RETRY_INTERVAL_SECONDS=300\`, \`BUNDLE_REPAIR_LARGE_RETRY_BATCH_SIZE=500\`, \`BUNDLE_REPAIR_LARGE_BUNDLE_THRESHOLD=50\`. Setting batch_size to 0 soft-disables the lane (loop iterates over empty array).

## Risk audit — what could go wrong

- **Data integrity**: zero risk. The new SQL is SELECT-only. The new index is partial + additive. SQLite atomically maintains index consistency with row writes.
- **Same bundle picked by both queries in one cycle**: \`txFetcher.queueTxId\` deduplicates against its in-memory queue snapshot — the second call is a no-op. The only observable effect is \`retry_attempt_count\` incrementing twice instead of once, which is cosmetic.
- **Brand-new uploads with no data_item_count**: only picked by the existing query, get downloaded normally, become eligible for the new lane after first unbundle. **Not starved.**
- **Threshold misconfigured to 1**: both queries pick the same population, doubling retry attempts. Still works.
- **Threshold misconfigured very high**: new query returns empty. Effectively disables the lane.
- **Migration cost on large DBs**: partial WHERE keeps the index tiny — on a 9.8M-row gateway with ~1k matching rows the index is a few MB and the CREATE takes ~10 s. ANALYZE adds another ~10 s.

## Verification

Migration syntax (fresh DB):
- ✓ up runs cleanly
- ✓ down runs cleanly
- ✓ up re-runs cleanly (idempotent)

EXPLAIN on a copy of a 9.8M-row live \`bundles.db\` after applying the migration:

**Existing query (unchanged behavior)**:
\`\`\`
SEARCH TABLE bundles USING INDEX bundles_active_retry_priority_idx (last_fully_indexed_at=?)
\`\`\`
PR #719's index still serves the existing query. No regression.

**New query**:
\`\`\`
SEARCH TABLE bundles USING INDEX bundles_large_active_priority_idx (data_item_count>?)
USE TEMP B-TREE FOR RIGHT PART OF ORDER BY
\`\`\`
Uses the new partial index. The "RIGHT PART" temp sort is the trailing \`retry_attempt_count, last_retried_at\` columns within each \`data_item_count\` value — unavoidable given DESC on the leading column. Acceptable: the candidate set is tiny (442 rows in our test).

**New query runtime on the live-data copy**: 13 ms (442 results).

## Stacks on

Based on \`perf/data-importer-memory\` (PR #720). Once that merges, GitHub will auto-rebase this onto develop.

## Test plan

- [x] SQL syntax verified on fresh DB
- [x] Migration up/down/idempotency verified
- [x] EXPLAIN verified — no regression on existing query, new query uses new index
- [x] New query runtime measured on production-scale DB copy (13 ms)
- [x] Race-condition design walked through (see risk audit)
- [x] Test schema updated (\`test/bundles-schema.sql\`) with the new index
- [ ] \`yarn test\` — local Node version mismatch with ESLint 9.37; deferred to CI
- [ ] Local deployment + observation of large-bundle retry rate after the fix

### DIFF
--- a/migrations/2026.05.13T18.00.00.bundles.add-large-bundle-priority-index.sql
+++ b/migrations/2026.05.13T18.00.00.bundles.add-large-bundle-priority-index.sql
@@ -1,0 +1,22 @@
+-- up migration
+--
+-- Adds bundles_large_active_priority_idx — a partial index supporting the
+-- new `selectLargeFailedBundleIds` SQL statement (parallel retry lane for
+-- Turbo-class bundles).
+--
+-- Partial WHERE predicate keeps the index tiny: it only covers bundles that
+-- are (a) not yet fully indexed and (b) known to have data_item_count >= 50
+-- (Turbo bulk threshold; direct ArDrive bundles have 1-2 DIs). On a 9.8M-row
+-- bundles.db with ~1k known-large bundles in flight, the index is a few MB.
+--
+-- The 50 literal in the partial WHERE matches the default value of the
+-- BUNDLE_REPAIR_LARGE_BUNDLE_THRESHOLD env var. If operators lower that env,
+-- the partial index will still serve queries (just for the >= 50 subset);
+-- queries with threshold < 50 will fall back to a full scan. Operators who
+-- want to lower the threshold permanently should also adjust this index.
+
+CREATE INDEX IF NOT EXISTS bundles_large_active_priority_idx
+  ON bundles (data_item_count, retry_attempt_count, last_retried_at)
+  WHERE last_fully_indexed_at IS NULL AND data_item_count >= 50;
+
+ANALYZE bundles;
--- a/migrations/down/2026.05.13T18.00.00.bundles.add-large-bundle-priority-index.sql
+++ b/migrations/down/2026.05.13T18.00.00.bundles.add-large-bundle-priority-index.sql
@@ -1,0 +1,2 @@
+-- down migration
+DROP INDEX IF EXISTS bundles_large_active_priority_idx;
--- a/src/config.ts
+++ b/src/config.ts
@@ -5,9 +5,10 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 import { canonicalize } from 'json-canonicalize';
-import { isMainThread } from 'node:worker_threads';
 import { existsSync, readFileSync } from 'node:fs';
+import { isMainThread } from 'node:worker_threads';
 
+import { verificationPriorities } from './constants.js';
 import { createFilter } from './filters.js';
 import * as env from './lib/env.js';
 import { initHttpSig, saveAttestationTxId } from './lib/httpsig.js';
@@ -15,9 +16,8 @@ import type {
   HttpSigObserverContext,
   HttpSigSignerContext,
 } from './lib/httpsig.js';
-import { release } from './version.js';
 import logger from './log.js';
-import { verificationPriorities } from './constants.js';
+import { release } from './version.js';
 
 //
 // HTTP server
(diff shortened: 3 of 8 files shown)


vilenarios:
@coderabbitai review

vilenarios:
Closing — instrumentation in #728 showed this PR is a no-op in practice.

After deploying #728 we observed:
- `data_importer_queue_full_skips_total{importer="bundleDataImporter"} = 572` in 21 min
- `bundleDataImporter` queue continuously at its 1000-item cap
- Zero downloads >10 MB completed in 21 min (turbo bundles)
- Effective throughput ~24 MB/min, network-limited, not scheduling-limited

`retryLargeBundles` was correctly scheduling turbo bundle retries, but every retry was silently dropped at `DataImporter.queueItem`'s queue-full check before reaching a worker. The actual bottlenecks are upstream txFetcher saturation and downstream network throughput. Smarter retry ordering doesn't help when the queues are saturated and the pipe is full.

Keeping the branch for now in case we revisit the design (e.g., paired with a dedicated worker pool or queue-size bump). Closing the PR so it doesn't sit open against develop as a stale recommendation.

PR #734, opened 2026-05-14
fix(stream): wall-clock cap on attachStallTimeout to prevent paused-stream wedge

vilenarios:
## Summary

`attachStallTimeout`'s per-data stall timer is cleared on every `pause` event so that legitimate backpressure does not look like an upstream stall. That leaves an edge case: if an upstream peer goes silent *while* the consumer is paused for backpressure, neither `data`, `pause`, nor `resume` fires again, the stall timer is never re-armed, and the stream hangs forever — no `end`, no `error`, no `close`.

In production this manifested as a creeping wedge of the bundle backfill pipeline. With `BACKGROUND_RETRIEVAL_ORDER=ar-io-network,...` the gateway ran fine for ~30 minutes, then bundle downloads stopped completing. No errors logged. `bundleDataImporter` queue stayed pegged at 1000/1000. CPU 40% (live traffic unaffected) but `bundle_download_duration_seconds_count` flatlined. Root cause: 24 download workers accumulated in this hung-paused state, holding `peerRequestLimiter` slots forever via `streamPeerCounts` because `stream.once('close', …)` never fires for a stream stuck in paused-with-no-data limbo.

## Fix

Adds an optional `maxRequestMs` to `attachStallTimeout` — a wall-clock timer that fires regardless of pause/resume state. Mirrors the existing `connectionTimer` idiom in `ar-io-data-source.ts:130` that already caps the connection phase. On timeout the stream is destroyed with a clear `wall-clock timeout` error, which:
- Propagates as a normal request failure
- Releases the limiter slot via existing `onRelease` / `close` handler path
- Lets the next retry pick a different peer

## Defaults

`STREAM_REQUEST_TIMEOUT_MS=900000` (15 min) — generous enough that legitimately slow transfers (~1.1 MB/s for a 1 GB bundle) survive. Operators tune via env if peers are reliably fast or massive transfers need more headroom.

## Scope of plumbing

- `ArIODataSource` and `GatewaysDataSource` both pass the new cap (both call `attachStallTimeout` on HTTP response streams that can backpressure against the cache write).
- `ans-104.ts` and `data-root.ts` also call `attachStallTimeout` but for internal streams that don't backpressure against a network peer — they keep the unchanged 2-arg form (backward compatible) until there's evidence they need the cap too.
- The new param is optional; calling sites without it behave exactly as before.

## Test plan
- [x] New unit tests cover: (a) destroy fires after `maxRequestMs` even when stream is paused throughout, (b) timer is cleared when stream ends naturally, (c) backward compat — no cap when arg omitted.
- [x] Existing `attachStallTimeout` tests still pass unchanged.
- [ ] CI: `npm test` for `lib/stream.test.ts`
- [ ] Local deploy + soak: confirm pipeline does not wedge after >30 min on `BACKGROUND_RETRIEVAL_ORDER=ar-io-network,...`, which was the original repro condition.


vilenarios:
Closing — superseded by #737.

When deployed, the wall-clock cap (`maxTimer`) never actually fired in the wedge condition. Diagnosis (with PR #736's phase counters): the bug was structural — `ReadThroughDataCache` returned the same `data.stream` to two consumers (the cache pipeline AND the caller), and pipeline's `'close'` event triggered `attachStallTimeout`'s cleanup which cancelled `maxTimer` before it could destroy a wedged stream.

PR #737 fixes the wedge at the source (tee the stream so cache + caller have independent backpressure). With #737 the safety net this PR added is no longer needed — the wedge is structurally impossible.

Code in this PR isn't harmful (the cap could still catch unrelated future bugs), but it didn't solve what we hoped and adds complexity for no observed benefit. Keeping the branch around for reference.

PR #740, opened 2026-05-15
perf(unbundle): skip re-emit for already-indexed children (PE-9098)

arielmelendez:
## Summary

When the bundle repair worker re-queues a bundle whose previous parse had some children dropped at the data-item indexer's queue cap, the parser used to re-emit **every** child — including the 90%+ already indexed. Each redundant emit consumed a slot in the `dataItemIndexer` / `ans104DataIndexer` queues, contending with the genuinely-new children for the cap. Under sustained backfill load this is the dominant contributor to the **2M+ items dropped per hour** we observed on the Turbo gw2 indexer after PE-9098 landed.

This PR adds a skip-set: before parsing, the `Ans104Unbundler` queries `BundleIndex.getIndexedChildIds(parentId)` for ids already in `new_data_items` ∪ `stable_data_items` under this parent. The set is passed through to the parser worker. For each parsed child whose id is in the set, the parser counts it toward `matchedItemCount` (preserving the bundle's `matched_data_item_count` invariant — see below) and skips the filter eval, hash, and `DATA_ITEM_MATCHED` emit.

## Why count skipped items toward `matchedItemCount`

`upsertBundle` updates `matched_data_item_count` with `IFNULL(@matched, matched)` — passing a value replaces it. The fully-indexed predicate is **strict equality** between `bundle_data_items` row count and `matched_data_item_count`. If we lowered the emitted matched count by the skip amount, the new lower value would replace the existing one, and the equality would never hold afterwards (rows > matched, so `fully_indexed_at` never sets and the bundle stays in the repair pool forever). Counting skips toward `matchedItemCount` avoids that.

Assumption: the unbundle/index filter has not changed between parses. When filters do change, the filter-reprocess path clears the bundle's state separately and `skipChildIds` will be empty for those bundles — so this assumption is enforced at the workflow layer, not implicit.

## Failure handling

If `getIndexedChildIds` throws (DB lock, breaker open, etc.) we log and proceed without a skip-set. Worst case is the pre-fix behavior of redundant emits — not a regression, not a hard failure.

## Test plan

- [x] `yarn build` clean
- [x] `yarn lint:check` clean on changed files
- [x] `yarn test:file src/workers/ans104-unbundler.test.ts src/workers/bundle-repair-worker.test.ts` — 12 tests pass
- [ ] Deploy to gw2 (which is already running the predecessor PR): expect `data_items_dropped_total{queue_name="dataItemIndexer"}` and `{queue_name="ans104DataIndexer"}` rates to drop dramatically (skip rate on repair-driven re-parses should also fall as queue pressure decreases). New `alreadyIndexedSkippedCount` is included in the `UNBUNDLE_COMPLETE` message; surfacing it as a metric is a follow-up.

## Notes

- Base branch is `PE-9098-bundle-repair-bdi-routing` (the previous PR). Once that lands, this PR's diff will be just the skip-set change.
- Five-edit DB plumbing for `selectIndexedChildIds` + `getIndexedChildIds`: SQL, worker impl, queue wrapper, message handler, `BundleIndex` interface.
- The new `alreadyIndexedSkippedCount` value flows out of the parser worker through `UNBUNDLE_COMPLETE` but isn't yet wired to a Prometheus counter. A small follow-up should add one so we can verify the optimization from Grafana.

### DIFF
--- a/src/database/sql/bundles/repair.sql
+++ b/src/database/sql/bundles/repair.sql
@@ -1,3 +1,16 @@
+-- selectIndexedChildIds
+-- Returns ids of all data items already indexed as children of @parent_id,
+-- across both new_data_items (pre-flush) and stable_data_items
+-- (post-flush). Used by the Ans104Unbundler before parsing to build a
+-- skip-set: any child the parser would otherwise emit but whose id is
+-- already in this set can be dropped pre-queue, avoiding redundant
+-- queue pressure and downstream upsert no-ops when a bundle is
+-- re-parsed because some prior children were dropped at the data-item
+-- indexer's queue cap.
+SELECT id FROM new_data_items WHERE parent_id = @parent_id
+UNION
+SELECT id FROM stable_data_items WHERE parent_id = @parent_id
+
 -- selectFailedBundleIds
 SELECT DISTINCT id
 FROM (
--- a/src/database/sql/bundles/repair.sql
+++ b/src/database/sql/bundles/repair.sql
@@ -1,15 +1,25 @@
 -- selectIndexedChildIds
--- Returns ids of all data items already indexed as children of @parent_id,
--- across both new_data_items (pre-flush) and stable_data_items
--- (post-flush). Used by the Ans104Unbundler before parsing to build a
--- skip-set: any child the parser would otherwise emit but whose id is
--- already in this set can be dropped pre-queue, avoiding redundant
--- queue pressure and downstream upsert no-ops when a bundle is
--- re-parsed because some prior children were dropped at the data-item
--- indexer's queue cap.
-SELECT id FROM new_data_items WHERE parent_id = @parent_id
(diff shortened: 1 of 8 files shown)


arielmelendez:
Might not ship this after all.

vilenarios:
Hey @arielmelendez — we're prepping a release and noticed your comment "might not ship this after all." Should we close PR #740, or do you want to revisit? Just want to align before cutting the next release. cc @vilenarios

arielmelendez:
> Hey @arielmelendez — we're prepping a release and noticed your comment "might not ship this after all." Should we close PR #740, or do you want to revisit? Just want to align before cutting the next release. cc @vilenarios

@vilenarios this hasn't been integrated or adequately tested yet and it's not clear to me that it's yet necessary for optimal system performance in the real-life scenarios we're encountering. Let's let this one hang back a bit longer while we continue fine tuning the system in other areas a bit more.

PR #747, opened 2026-05-21
feat(data-item-indexer): make worker count env-configurable

vilenarios:
## Summary

`DataItemIndexer`'s fastq concurrency is hardcoded to `1` — every `saveDataItem` call awaits before the next starts. Under heavy backlog drain (failed bundles with 50-500 items each) the indexer becomes the throughput bottleneck at ~40K items/hr because the main thread can't pipeline JS prep (event emit, metric inc, log) for the next item while the prior `saveDataItem` is in flight to the SQLite worker thread.

The `DataItemIndexer` constructor already accepts a `workerCount` option; this PR just plumbs it through via a new env var.

## Why now

Observed on this gateway during a 2.78M failed-bundle drain cycle: peak rate plateaued at ~40K data-items/hr regardless of upstream rate, with `sqlite_in_flight_ops` snapshots showing the SQLite write worker mostly idle (suggesting the bottleneck is the *serial-await* pattern in the dispatcher, not the SQL execute itself). Modern small-bundle load is unaffected because ~2 items/bundle keeps the indexer trivially caught up. Large failed-bundle retries with ~85 items/bundle saturate it.

## Changes

- Add `DATA_ITEM_INDEXER_WORKER_COUNT` env var (default `1`, backward compatible)
- Pass `config.DATA_ITEM_INDEXER_WORKER_COUNT` to the constructor in `system.ts`
- docker-compose env passthrough
- `docs/envs.md` entry

## Bounding the upside

SQLite is single-writer per database, so the upper bound on speedup is the gap between per-item JS work and per-item SQLite execute. At observed ~90ms end-to-end per item, there's likely headroom. Try 4 first; back off if main-thread CPU saturates.

## Test plan

- [x] tsc clean
- [x] prettier clean
- [x] eslint clean
- [x] No code path changes — only constructor param wiring (existing tests cover the param)
- [ ] CI: full suite
- [ ] Deploy + soak with `DATA_ITEM_INDEXER_WORKER_COUNT=4` set in `.env`; measure indexer throughput during BRW backlog drain vs current ~40K items/hr ceiling


vilenarios:
@coderabbitai review

PR #783, opened 2026-06-18
feat(chunk): optimistic chunk ingest caching (POST /chunk write-through) [WIP]

vilenarios:
> **Status: feature-complete, draft until soak.** Off by default (`CHUNK_INGEST_CACHE_ENABLED=false`) — inert in production until enabled. Build + unit tests + typecheck (`tsconfig.prod.json`) + lint all green locally.

## What & why

Today `POST /chunk` broadcasts a chunk to the Arweave network and **discards the bytes locally** — if anyone later reads that data, the gateway re-fetches it over the network. This makes the gateway **optimistically cache the chunks posted through it**, so data uploaded via a gateway (e.g. a baremetal Turbo bundler on `turbo-gateway.com`) is immediately servable from that node.

- **Verify-first:** each posted chunk is validated against its asserted `data_root` (`validateChunk`) before caching — cached `verified`-by-construction.
- **Safe open ingest:** merkle binding means a *real* transaction cannot be poisoned. A chunk under a `data_root` that never confirms on-chain is **unaddressable** (no GET-by-`data_root` endpoint) and is reclaimed by GC. Optional `CHUNK_INGEST_CACHE_ALLOWLIST` gates *caching* (not posting).
- **Confirmation pushed from indexing:** `TX_INDEXED` → `confirmChunkPlacements(data_root)`; no `data_root→tx` index needed. Unconfirmed placements ride a tiered TTL and are GC'd.

## Design

`chunk_placements` (new `chunks.db`) is both the optimistic ingest ledger (`origin`/`cached_at`/`confirmed_at`) and a chunk index keyed by `(data_root, relative_offset)`.

**Absolute-offset reads already work without new read-path code.** An ingest chunk is keyed `(data_root, relative_offset)`; `ChunkRetrievalService` resolves a `GET /chunk/{absOffset}` through `TxBoundarySource` → `(data_root, relative_offset)` → the read-through cache hits on that key. The tx boundary is resolvable *exactly when the tx is on-chain* = when the chunk is confirmed. So the deeper #5a "replace FS-as-index with SQLite / retire symlinks / migrate the cache" refactor was **intentionally not done** here — it's a separable hot-read-path change not required for the goal.

## Done

- [x] `chunks.db` migration + `chunk_placements` schema + statements
- [x] 5th DB wired through the `StandaloneSqlite` worker pool; placement methods
- [x] `POST /chunk` verify + write-through cache (fire-and-forget)
- [x] allowlist / enabled gating + config + metrics
- [x] confirmation subscriber on `TX_INDEXED`
- [x] GC worker (tiered TTL + runaway disk-cap) + store `del()`
- [x] confirmation-latency histogram (to tune the timeout defaults from soak data)
- [x] `docs/envs.md` + `docker-compose.yaml` for the 8 env vars
- [x] unit tests (GC sweep, ingest rejection) — 6/6 pass

## Defaults (reasoned; timeouts to be finalized from soak data)

- Confirmation TTL **6h** open / **24h** allowlisted (erring long: too-short only loses cache benefit; too-long is bounded by the disk-cap and unserveable).
- `CHUNK_INGEST_MAX_PENDING_BYTES` defaults **25 GiB** (non-zero so enabling open ingest can't silently fill the disk; the TTL is the primary reclamation, this is a runaway backstop).

## Remaining before merge

- [ ] E2E (POST a real chunk → served by id and by absolute offset)
- [ ] Soak on a live gateway; read `chunk_ingest_confirmation_latency_seconds` and tune the TTLs

## Metrics

`chunk_ingest_cache_total{result}`, `chunk_ingest_confirmed_total`, `chunk_ingest_confirmation_latency_seconds`, `chunk_ingest_evicted_total{reason}`, `chunk_ingest_pending_bytes`.


vilenarios:
## Integrity note for reviewers — the permanence invariant

This feature must never serve not-yet-permanent data as verified/permanent, and must reclaim anything that never mines. How that holds in this PR:

- **Cleanup is confirmation-driven, not blind TTL.** A placement is kept iff its `data_root` gets indexed (`TX_INDEXED` → `confirmChunkPlacements`); never-confirmed placements are evicted by the GC sweep. The TTL is just the patience window.
- **Pending chunks are structurally unserveable here.** Every serving path resolves `id/offset → data_root` through the index, which only contains mined (confirmed) txs; the retrieval-hint headers carry `rootTxId`, not `data_root`. So there is no reachable path that serves an unconfirmed ingest chunk.
- **`verified` requires real verification** (the DB column set by the verification worker after confirmation). Optimistic/pending data is never labeled verified.

The active "don't-serve-until-`confirmed_at`" read guard is **deferred to the follow-on optimistic-tx-index feature**, because optimistic tx indexing is what would make a pending tx's `data_root` resolvable — i.e. that feature *creates* the exposure, so the guard is gated there as a blocking acceptance criterion.

vilenarios:
## Route-P validation — consumption proven ✅

Validated end-to-end on a live, fully-indexing gateway at `5c75a917` (off-by-default; deployed feature-on-latest, reverted cleanly, production unaffected):

- **The cache is consumed.** A confirmed, ingest-cached chunk is **served and unbundled from local cache with zero network re-fetch** — 13 `ReadThroughChunkDataCache` hits for the bundle's `data_root`, with the contiguous cache **cold** (so this is the chunk path, not a contiguous-cache masking artifact). **Negative control:** same window, **2266** network chunk fetches for non-ingest-cached `data_root`s vs **0** for the cached bundle.
- **The orphaned-cache bug is fixed.** Pre-`5c75a917`, ingest keyed chunks by the Arweave END offset while every read path queries the START offset → every lookup missed → S3/network fallthrough. `5c75a917` keys storage by START → hits.
- **GC + disk bounded.** `chunk_ingest_pending_bytes` returns to 0 (confirmed chunks leave the pending pool); the GC worker evicts the never-confirmed tail.
- **No indexing regression** — the `chunks.db` wiring runs regardless of the flag; indexer lag ≈ 0, GraphQL normal, no errors.
- **Real confirmation latency** ~2–7 min (small sample). The 6h/24h TTL defaults are intentionally conservative and retained; tightening would need a sustained multi-hour distribution — operational, not a correctness gate.

This is the round-trip that the isolated unit tests structurally couldn't cover, caught by the soak.

PR #789, opened 2026-06-19
feat(parquet-export): spill DuckDB sorts to disk to survive dense partitions

arielmelendez:
## Problem

`parquet-export` (run in the **core** service via `/ar-io/admin/export-parquet`, driven by the ClickHouse auto-import) fails on dense, bundle-heavy height partitions for **two independent reasons**, and the real cause was hidden behind a generic worker exit code. On the canary gateways this left a large `stable_data_items` residual that could never export → never import → never prune, so SQLite never drained and `minStableDataItem` stayed pinned — the ClickHouse flush was effectively stalled.

### Failure 1 — in-memory sort OOM
The exporter runs `COPY (... ORDER BY ...) TO parquet` inside an in-memory DuckDB with no `temp_directory`, so the sort runs entirely in RAM. A large partition (millions of rows plus their tags) exhausts memory and the worker is OOM-killed. `maxFileRows` `LIMIT/OFFSET` pagination doesn't help — each page re-sorts the full set.

### Failure 2 — `root_parent_offset` UINT32 overflow
The export cast `root_parent_offset::UINTEGER` (UINT32), but it is a byte offset within the root L1 transaction and exceeds 4 GiB (>UINT32) for data items in large bundles. DuckDB throws `Conversion Error … out of range for the destination type UINT32` during the COPY, aborting the export of any partition containing such a row. Every other byte-offset column on the same SELECT was already `UBIGINT`; this one was the outlier, and ClickHouse already stores the column as `UInt64`.

Both failures surfaced only as `Worker stopped with exit code 1`, because the worker's `exit` handler overwrote any error that had already been reported.

## Changes
1. **Spill DuckDB sorts to disk** — set `temp_directory` + `memory_limit` (plus `max_temp_directory_size`, `preserve_insertion_order=false`, and an optional `threads` cap) on the export connection so large sorts process out-of-core instead of OOM-killing the worker. New env vars (`PARQUET_EXPORT_DUCKDB_MEMORY_LIMIT`, `PARQUET_EXPORT_DUCKDB_MAX_TEMP_DIRECTORY_SIZE`, `PARQUET_EXPORT_DUCKDB_THREADS`), threaded through to the worker. The pragma helper is extracted and unit-tested against a real DuckDB connection.
2. **Widen the `root_parent_offset` cast** `::UINTEGER` → `::UBIGINT` to match the other byte-offset columns and the ClickHouse `UInt64` column. Non-destructive: source SQLite is read-only, ClickHouse is already `UInt64`, no migration — only the parquet column widens UINT32 → UINT64. Existing UINT32 files remain valid and import into the `UInt64` column (one-file-at-a-time import coerces both widths).
3. **Surface the real export error** — guard the terminal handlers with a `settled` flag so the `exit` handler no longer clobbers a specific error with a generic message, and create the worker with `stderr: true` to capture its stderr, so a native crash that kills the worker before any JS catch (e.g. an OOM abort) reports the real cause instead of a bare exit code. (This is what revealed Failure 2.)
4. **Wire the DuckDB limit env vars into the compose `core` service** so they can be tuned at runtime, keeping `docker-compose.yaml` and `docs/envs.md` in sync per CLAUDE.md.

## Testing
- `tsc --noEmit` (prod project) and `eslint src test`: clean
- New unit test for the DuckDB-pragma helper: passes
- Auto-verify adapters already project `root_parent_offset` via `Number()` into the canonical `number` shape, so the UINT32 → UINT64 widening needs no adapter change.
- **Validated end-to-end on the canary gateways:** after deploy, the previously-failing dense partitions export cleanly with zero cast errors. Both gateways fully drained their `stable_data_items` residual — one cleared a ~14.5M-row bottom bucket in a single default-interval (10000-height) window — with `minStableDataItem` advancing to the chain tip and the ClickHouse flush resuming.

## Deploy note
The export runs in the **core** service, so this requires a core image rebuild/deploy, not just an auto-import restart.


arielmelendez:
Thanks @coderabbitai. On the `src/routes/ar-io.ts:34-44` finding (move `ParquetExporter` into `src/system.ts` DI):

The `getParquetExporter()` lazy singleton is **pre-existing** — this PR only threads three new DuckDB-limit args into the existing `new ParquetExporter(...)` call and doesn't change *where* the service is constructed (it's flagged as outside the diff range for that reason).

Relocating its lifecycle into `system.ts` is a reasonable guideline-alignment, but it's a separate refactor: it would also flip the current lazy, construct-on-first-export-request pattern (which self-registers its own cleanup handler in the route) to eager boot-time construction. That's out of scope for this targeted OOM fix, so I'm deferring it to a follow-up rather than expanding this PR.

PR #799, opened 2026-06-26
fix(chunks): reject and self-heal zero-length chunk cache poison

arielmelendez:
## Problem

A data request served from chunks produced a Honeycomb trace with **5.1M spans** — `RawDataHandler.handle` → `ReadThroughDataCache.getData` → a seemingly endless waterfall of `ReadThroughChunkDataCache.getChunkDataByAny`, all at the **same relative offset** and almost all sub-millisecond cache hits. The request ran until the 15s HTTP timeout.

The transaction was a **2805-byte, single-chunk L1 tx** — it should be served in one chunk and stop.

## Root cause

`TxChunksDataSource`'s full-stream read loop terminates only on byte accounting (`bytes < size`), advancing by `chunkData.chunk.length`. A **zero-length chunk** never advances `bytes`, so `bytes < size` stays true forever and the loop re-requests the **same offset** indefinitely.

The zero-length chunk came from a **poisoned cache entry**: `FsChunkDataStore` wrote chunk bytes with no length validation, so a source that returned an empty chunk once was persisted as a **0-byte file**. Thereafter `has()` reports a hit on that file and `get()` serves an empty buffer — re-poisoning every subsequent read of that offset. The first span in the trace confirmed the very first read was already a `source: cache` hit returning empty.

## Fix — defense in depth across every layer

- **`FsChunkDataStore.set`** — refuse to persist a zero-length chunk (also guards the `/chunk` ingest path).
- **`FsChunkDataStore.get` / `getByAbsoluteOffset`** — treat an existing 0-byte file as a **miss**, so already-poisoned entries **self-heal**: the caller refetches and overwrites the entry on next access.
- **`ReadThroughChunkDataCache`** — reject a zero-length chunk returned by the source (never cache it; throw so the retrieval cascade falls through to another source).
- **`TxChunksDataSource`** — forward-progress guards in the stream loop: abort on a non-advancing (zero-length) chunk, and abort on a chunk-count overrun (`> ceil(size / MAX_CHUNK_SIZE) + 1`) as a backstop against pathological tx geometry. Gated on `size > 0`.

## Observability

New metric `chunk_zero_length_total{stage}` counts rejected zero-length chunks by pipeline stage (`source_fetch` / `cache_read` / `cache_write`). `TxChunksDataSource` aborts increment `chunk_stream_aborts_total{reason}`.

## Notes

- Existing poisoned cache entries self-heal on next access; no manual eviction required.
- Updated a prior `FsChunkDataStore` test that asserted empty chunks round-trip — that was effectively the poison contract.
- **Not addressed here (follow-up):** the per-chunk span instrumentation emits one span (plus children) per chunk read with no per-request cap or sampling, which is what let a runaway loop produce a multi-million-span trace. The guards prevent the runaway, but large legitimate files still create large traces.

## Testing

- `src/store/fs-chunk-data-store.test.ts`, `src/data/read-through-chunk-data-cache.test.ts`, `src/data/tx-chunks-data-source.test.ts` — added coverage for set-rejection, read self-heal, source-rejection, the zero-length guard, and the count cap. All pass.
- `yarn lint:check` and `yarn build` clean.

### DIFF
--- a/src/data/read-through-chunk-data-cache.test.ts
+++ b/src/data/read-through-chunk-data-cache.test.ts
@@ -115,5 +115,33 @@ describe('ReadThroughChunkDataCache', () => {
       assert.deepEqual(chunkDataStoreSetSpy.mock.callCount(), 1);
       assert.deepEqual(networkSpy.mock.callCount(), 1);
     });
+
+    it('should reject a zero-length chunk from the source without caching it', async () => {
+      const chunkDataStoreGetSpy = mock.method(
+        chunkDataStore,
+        'get',
+        async () => undefined,
+      );
+      const chunkDataStoreSetSpy = mock.method(chunkDataStore, 'set');
+      mock.method(chunkSource, 'getChunkDataByAny', async () => ({
+        hash: Buffer.alloc(0),
+        chunk: Buffer.alloc(0),
+      }));
+
+      await assert.rejects(
+        () =>
+          chunkCache.getChunkDataByAny({
+            txSize: TX_SIZE,
+            absoluteOffset: ABSOLUTE_OFFSET,
+            dataRoot: B64_DATA_ROOT,
+            relativeOffset: 0,
+          }),
+        { message: /zero-length chunk/ },
+      );
+
+      assert.deepEqual(chunkDataStoreGetSpy.mock.callCount(), 1);
+      // The invalid empty chunk must not be persisted.
+      assert.deepEqual(chunkDataStoreSetSpy.mock.callCount(), 0);
+    });
   });
 });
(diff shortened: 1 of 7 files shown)


arielmelendez:
Thanks @coderabbitai — addressed both nitpicks in 3f331cb3:

- **`src/metrics.ts`** — added TSDoc to `chunkStreamAbortsTotal` and `chunkZeroLengthTotal`, documenting each label value (`reason`: `zero_length_chunk` / `chunk_count_exceeded`; `stage`: `source_fetch` / `cache_read` / `cache_write`).
- **`src/store/fs-chunk-data-store.test.ts`** — added two cases for the absolute-offset path: a pre-existing 0-byte `by-absolute-offset` entry reads as a miss via `getByAbsoluteOffset()`, and a zero-length `set(..., absoluteOffset)` creates no absolute-offset index.

All store tests pass (17/17); `yarn lint:check` clean.

PR #804, opened 2026-06-29
fix(chunk-post): key sorted-peer memo by set, not length

vilenarios:
## What

`ArweaveCompositeClient.getSortedChunkPostPeers` memoizes its weighted peer ordering with a normalizer of `eligiblePeers.length`. Two **different** eligible-peer sets of equal size therefore share one cached ordering for the full `CHUNK_POST_SORTED_PEERS_CACHE_DURATION_MS` window (10s default).

This keys the memo by the **set of peer ids** instead (order-independent `slice().sort().join()`).

## Why it matters

The length-only key quietly narrows chunk-POST **seeding diversity**: a churning-but-same-size peer set keeps reusing the first set's ordering, so the same peers get prioritized for up to 10s even as the eligible set changes. For a gateway whose job is to be a good seeding *target*, fanout diversity is exactly what we don't want to silently pin.

Surfaced during an architecture review of chunk fanout (the broader durability work concluded the gateway should optimize *fanout quality*, not duplicate the bundler's durability engine — this is one of those quality fixes).

## Cost

Still cheap: one sort+join over the bounded `postChunk` peer set (tens–low hundreds of entries), far less than the weighted `selectPeers` + filtering the memo guards. The 10s `maxAge` still absorbs gradual weight drift. Does not mutate the caller's array.

## Tests

Extracted the keying as the pure, exported `chunkPostPeersCacheKey` and unit-tested it directly:
- distinct same-length sets ⇒ distinct keys (the bug)
- order-independent for the same set
- no input mutation
- empty set

`yarn test:file src/arweave/composite-client.test.ts` → 15/15 pass; eslint clean.

### DIFF
--- a/src/arweave/composite-client.test.ts
+++ b/src/arweave/composite-client.test.ts
@@ -8,7 +8,10 @@ import { strict as assert } from 'node:assert';
 import { describe, it, beforeEach, afterEach, mock } from 'node:test';
 import { default as Arweave } from 'arweave';
 
-import { ArweaveCompositeClient } from './composite-client.js';
+import {
+  ArweaveCompositeClient,
+  chunkPostPeersCacheKey,
+} from './composite-client.js';
 import { UniformFailureSimulator } from '../lib/chaos.js';
 import { ArweavePeerManager } from '../peers/arweave-peer-manager.js';
 import log from '../log.js';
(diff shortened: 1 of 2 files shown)


vilenarios:

PR #805, opened 2026-06-29
feat(ip-utils): add ipFaultDomain for peer network-block bucketing

vilenarios:
## What

Pure, unwired additions to `src/lib/ip-utils.ts`:

- **`ipFaultDomain(host, { v4Bits=24, v6Bits=48 })`** — reduces a peer host to a stable "fault domain" bucket key (the network block it belongs to):
  - IPv4 (incl. IPv4-mapped IPv6) → masked to `/24` → `"a.b.c.0/24"`
  - IPv6 → expanded + truncated to `/48` → `"2001:db8:abcd::/48"`
  - non-IP (unresolved hostname) → returned verbatim (counts as its own domain)
- **`expandIpv6(ip)`** — expands to 8 zero-padded hex groups (handles `::`); the minimum IPv6 parsing needed for prefix bucketing (the existing `isIpInCidr` is IPv4-only).

## Why

**PR B1** of the chunk-fanout SPOF fix (design: `architect-handoff/chunk-seeding-durability-plan.md` §17). The five `tip-*.arweave.xyz` preferred chunk-POST nodes all resolve into **one `38.29.227.0/24`** — so today's hard `CHUNK_POST_MIN_PREFERRED_SUCCESS_COUNT=2` quorum is a single-fault-domain SPOF: a tips outage hard-fails every chunk POST even when independent discovered peers accepted it.

`ipFaultDomain` is the primitive the upcoming **B2** (soft-preferred quorum + distinct-domain counting) uses to detect that condition and route around it. The chunk-POST peer list is IP-literal after the peer manager's DNS resolution, so bucketing is in-process with no hot-path DNS.

## Behavior change

**None.** Pure helper additions, not yet called anywhere. Ships first so B2 is a smaller, focused review.

## Tests

`yarn test:file src/lib/ip-utils.test.ts` → 50/50 pass. New cases: the five tips collapse to one `/24`; distinct `/24`s stay distinct; IPv6 `/48` bucketing (same /48 ⇒ same bucket, different ⇒ different); IPv4-mapped normalization; custom prefix widths; `expandIpv6` full/compressed/malformed; unresolved-hostname fallback.

### DIFF
--- a/src/lib/ip-utils.test.ts
+++ b/src/lib/ip-utils.test.ts
@@ -15,6 +15,8 @@ import {
   isIpInCidr,
   isAnyIpAllowlisted,
   isAnyIpBlocked,
+  expandIpv6,
+  ipFaultDomain,
 } from './ip-utils.js';
 
 describe('IP Utilities', () => {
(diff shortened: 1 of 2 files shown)


vilenarios:

PR #806, opened 2026-06-29
feat(chunk-post): fault-domain-aware quorum + soft-preferred fallback (B2)

vilenarios:
**Stacked on #805** (`ipFaultDomain`) — base will retarget to `develop` once #805 merges.

## What

Makes chunk fanout **fault-domain-aware** and removes the **single-fault-domain tips SPOF**: all five `tip-*.arweave.xyz` resolve into one `38.29.227.0/24`, so today's hard `CHUNK_POST_MIN_PREFERRED_SUCCESS_COUNT=2` means a tips outage **hard-fails every chunk POST** even when independent discovered peers accepted it. Both new behaviors are **opt-in and default to the exact legacy quorum**.

## Changes

- **Distinct-domain counting** — buckets each successful peer by IP `/24`·`/48` (`ipFaultDomain`). `CHUNK_POST_MIN_DISTINCT_DOMAINS` (default `0` = off) adds a **soft** target among successes that degrades gracefully when the eligible peer set can't supply N (never hard-fails on scarcity).
- **Soft-preferred fallback** — `CHUNK_POST_PREFERRED_SOFT_FALLBACK` (default `false`) lets a strong distinct-domain discovered quorum satisfy a POST **only when no tip was eligible** (tips down/over-queue) — never when tips merely failed, so tips stay the hard primary on a real outage only.
- **Centralized verdict** — `evaluateChunkBroadcastVerdict` is the single source of truth for "seeded sufficiently?"; the handler reads `result.succeeded` instead of recomputing thresholds (removes latent drift). **With both flags at defaults it reduces *exactly* to the legacy rule** — covered by a regression test.
- **Placement-evidence contract (plan §18)** — `BroadcastChunkResult` now carries `distinctDomainCount` / `preferredEligibleCount` / `succeeded`, and `POST /chunk` emits the `X-AR-IO-Chunk-Placement-Domains` header, so the bundler can consume *where a chunk landed* instead of collapsing the result to pass/fail. (Bundler-side consumption is a separate cross-repo proposal.)
- **Metrics** — `chunk_post_distinct_domains` (histogram), `chunk_post_preferred_shortfall_total{reason}` (emitted even when the fallback makes the POST succeed → a tips outage is never silently masked), `chunk_post_domain_shortfall_total{reason}`.

## Safety / rollout

- **Default-off**: `MIN_DISTINCT_DOMAINS=0` + `PREFERRED_SOFT_FALLBACK=false` ⇒ byte-identical to today's behavior (regression-tested).
- Peer→domain bucketing is in-process (no hot-path DNS — the postChunk peer list is IP-literal after DNS resolution).
- Soak gate before any default flip: a simulated tips outage now *succeeds via fallback*, `chunk_post_distinct_domains` rises, no regression with flags off.

## Tests

`config` 20/20 · `composite-client` 24/24 (verdict logic exhaustively: default-off regression, fallback fires only on `tips_unavailable`, domain soft-degrade) · `ip-utils` 50/50 · lint clean.


vilenarios:

PR #808, opened 2026-06-30
test(e2e): stabilize the e2e harness (network isolation, reliable teardown, CI timeouts)

vilenarios:
## Summary

Fixes the **E2E harness instability** that has been reddening post-merge `develop`
runs of `build-core` — repeatedly, on *different* infra symptoms, with no relation to
the code under test. Most recently a single flake **hung for ~37 minutes** before being
cancelled by hand.

This is a test-infrastructure PR only — no production code changes.

## Root cause

`yarn test:e2e` runs all e2e suites **serially in one process**, and every suite stands up
the **production `docker-compose.yaml`**, which pins a single docker network name
(`DOCKER_NETWORK_NAME`, default `ar-io-network`) and publishes a fixed host port (`4000`).
When one suite's teardown was slow, partial, or threw, the next suite's `up()` collided:

```
a network with name ar-io-network exists but was not created for project "testcontainers-…"
Bind for 0.0.0.0:4000 failed: port is already allocated
```

…cascading into every subsequent suite (`test did not finish before its parent and was
cancelled`) and, in the worst case, hanging — and the job had **no timeout**, so a hang ran
until GitHub's 360-minute default.

A second contributing factor: **e2e only runs on `develop` + ubuntu, never on PRs**
(`build-core.yml` gates the step on `github.ref == 'refs/heads/develop'`), so this class of
failure can't be caught pre-merge.

## The fix — three layers

**Layer 0 — bound the runtime (CI).** Add a 30-min job timeout and a 20-min cap on the E2E
step in `build-core.yml` and `test-core.yml`. A wedged harness now fails fast instead of
burning ~37 min. Backstop only.

**Layer 1 — reliable teardown.** Replace every bare `compose.down()` with a shared
`composeDown()` (in `test/end-to-end/utils.ts`) that bounds the stop (`timeout`), removes
volumes so no state leaks between suites, tolerates a missing environment, and
swallows+logs teardown errors so a failed stop never masks a test result or aborts sibling
suites. Also bound startup (`withStartupTimeout`) so a wedged core fails fast.

**Layer 2 — network isolation.** Give each compose environment a unique
`DOCKER_NETWORK_NAME` (`ar-io-e2e-<pid>-<n>`). The compose file already parameterizes the
network name and services reference it by compose *key* (not literal name), so this is
transparent to inter-container DNS and to the tests (which reach the gateway over the host
port). A leaked or mid-teardown environment can no longer block the next suite. Bonus: e2e
no longer collides with a real `ar-io-network` on a developer's machine.

`bundler-sidecar.test.ts` has its own inline `composeUp`; it gets the same unique-network
treatment and uses `composeDown`.

## Deliberately *not* done (and why)

**Ephemeral host ports** (replacing fixed `4000` with `getMappedPort`, threaded through every
`localhost:4000` in the suites) would be the only way to run e2e *in parallel*. But the suite
runs serially (`--test-concurrency 1`), so port 4000 is exclusive once teardown reliably
releases it (Layer 1). The change would touch ~100 call sites across 7 files for no benefit
under serial execution — high risk, no payoff. Left as a future option if e2e is ever
parallelized.

## Verification

- `yarn lint:check` clean; no net-new `tsc` errors (the e2e tree has 10 pre-existing
  strict-mode errors on `develop`; this branch still has exactly 10).
- **No test assertions change** — suites still reach the gateway on the fixed host port.
- E2E doesn't run on PRs, so CI here covers build/lint/unit. To validate the harness change
  itself, dispatch the `test-core` workflow against this branch (it runs e2e on ubuntu
  unconditionally) — that's the real proving ground before merge.

### DIFF
--- a/.github/workflows/build-core.yml
+++ b/.github/workflows/build-core.yml
@@ -9,8 +9,10 @@ jobs:
         os: [ubuntu-latest, macos-latest]
     runs-on: ${{ matrix.os }}
     # Backstop against a hung E2E harness (a leaked compose env can otherwise
-    # wedge the run until GitHub's 360-minute default). Normal run is ~18 min.
-    timeout-minutes: 30
+    # wedge the run until GitHub's 360-minute default). Sized above the docker
+    # build (~13 min) plus the 20-min E2E step cap so the step timeout is what
+    # bounds a hang, not the job; the E2E suite itself runs ~10-15 min.
+    timeout-minutes: 45
     permissions:
       checks: write
       contents: read
--- a/.github/workflows/build-core.yml
+++ b/.github/workflows/build-core.yml
@@ -8,6 +8,9 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest]
     runs-on: ${{ matrix.os }}
+    # Backstop against a hung E2E harness (a leaked compose env can otherwise
+    # wedge the run until GitHub's 360-minute default). Normal run is ~18 min.
+    timeout-minutes: 30
     permissions:
       checks: write
       contents: read
@@ -37,6 +40,9 @@ jobs:
         run: docker build -t core .
       - name: Run E2E tests
         if: github.ref == 'refs/heads/develop' && matrix.os == 'ubuntu-latest'
+        # Tighter cap on just the E2E step so a stuck suite fails fast instead
+        # of consuming the whole job budget.
+        timeout-minutes: 20
         run: yarn test:e2e
         env:
           TEST_SKIP_TAGS: flaky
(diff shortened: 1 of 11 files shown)


vilenarios:
Closing — incomplete, abandoning the branch.

What it did accomplish: eliminated the multi-hour E2E hangs (CI timeouts + quarantining an infinite-loop test) and proved out docker network isolation. What it did NOT: make E2E green. The port-4000 cascade persists — the container-reap used the wrong label filter (`org.testcontainers` isn't applied to docker-compose-created containers) and never fired — and skipping one chronically-failing test only surfaced the next network-dependent timeout.

Root finding: E2E has been chronically red on develop due to multiple environment-dependent tests (data verification, bundle indexing — they fetch real data from trusted gateways, which doesn't complete in CI) on top of a shared host-port harness that cascades when any suite's teardown is skipped. This predates #807 and is unrelated to it. A real fix is a deliberate larger effort (ephemeral host ports / harness rework + flaky-test triage), not a reactive patch — so abandoning this branch rather than half-landing it.

PR #815, opened 2026-07-04
fix(chunk-ingest): make optimistic chunk confirmation sticky per data_root

vilenarios:
## Problem

Bundles our own AR.IO Bundler uploads are "findable-late" and 404 on `/raw`, and the indexer's unbundler fails for a sustained window — chronically for large bundles. On **gw1** (turbo-gateway.com, where the bundler posts at scale) the failing **2 GB** bundle `zwpirOGSpS_AEs6DxxV68lXgk8_4ohTrvEVH_dCtrTc` had only **~12% of its chunks on disk, gappy, with `relative_offset 0` missing**. Retrieval range-streams from offset 0 → immediate miss → peers (which don't yet have the fresh bundle) → fail. The 315 "failures" were ~9 such large bundles retried every 20 min.

This is a **retention/confirmation** defect, not a retrieval-addressing one. (A separate investigation ruled out the retrieval path: a hermetic repro showed the gateway already serves ingested chunks locally by `(data_root, relative_offset)` when they are present.)

## Root cause

`confirmChunkPlacements` is a **one-shot `UPDATE`** fired by the `TX_INDEXED` event (`system.ts`): it only sets `confirmed_at` on placements that exist **at that instant**. A multi-GB bundle streams its thousands of chunks in over a window far longer than that single event, so most arrive **after** it, never get confirmed, and are TTL-evicted by the ingest GC. Metrics on gw1: `chunk_ingest_evicted_total{reason="ttl"}=4000` vs `chunk_ingest_confirmed_total=543`.

## Fix

Make confirmation **sticky per `data_root`** (statement + one small table; no retrieval-path change, no hot-path tx lookup):

- **`saveChunkPlacement`** now inherits `confirmed_at` when the `data_root` is already confirmed, so every chunk ingested *after* the confirm event self-confirms at ingest.
- New **`confirmed_data_roots`** marker table, populated by `confirmChunkPlacements` — **gated on `EXISTS` in `chunk_placements`** so it stays bounded by *ingested* bundles rather than the whole tx index. The marker is (a) the primary inheritance source and (b) **excluded by the GC TTL sweep**, so a confirmed bundle is never partially evicted regardless of per-row state.

Net effect: once the one-shot `TX_INDEXED` confirms the first batch of a bundle, the entire bundle (present + all later-arriving chunks) is confirmed and retained → servable and unbundlable without waiting for network re-propagation.

## Correctness / scope

- **Bounded table:** the `EXISTS` gate means only data_roots we actually ingested chunks for are marked.
- **No new leak class:** confirmation still means "the tx was indexed"; confirmed chunks were already retained as the metadata index. This fix extends that to the *whole* bundle, which is the intent.
- **Retrieval unchanged.**
- **Residual (documented):** a chunk that arrives *and* expires before the very first confirm event still evicts — only reachable if the gateway's block-import lags beyond the ingest TTL (a sync pathology), not the observed failure.

## Tests

`src/database/chunk-placements.test.ts`:
- inherits `confirmed_at` from a confirmed sibling at ingest;
- **regression:** confirm a batch, then ingest more → auto-confirmed **and** not TTL-selected;
- marker shields a `data_root` from eviction even when a row is pending;
- marker `EXISTS`-gate: an indexed tx we never ingested chunks for is not marked and grants no phantom inheritance;
- explicit `confirmed_at` still wins.

All green (chunk-placements + ingest + GC + sql-loader + standalone-sqlite = 117/117); lint clean; migration up/down verified on a scratch DB; `test/chunks-schema.sql` regenerated.

## Deploy notes

One additive migration (`confirmed_data_roots`), no data rewrite; the SQL statement change takes effect on rebuild. On-by-default.


vilenarios:
## End-to-end validation on a real bundler → gateway pipeline

Deployed this branch to a live gateway (chunk ingest on, allowlisted bundler) and pushed **256 MB / 1025-chunk** bundles through it. Used a deliberately short ingest TTL (**180 s**, GC every **15 s**) so unconfirmed chunks actually evict within the test window — otherwise the default 2–24 h TTL makes any short test pass trivially.

### Run 1 caught a real bug (now fixed in 815b7017)

The confirming `TX_INDEXED` event fired **~2 s before** the bundler seeded the first chunk. This is the *dominant* ordering, not an edge case: the bundler's chunk-broadcast TX-confirmation gate holds seeding until the tx confirms network-wide, and the gateway imports that same block first. The marker's original `EXISTS in chunk_placements` gate found zero chunks → **no marker set** → all 1025 chunks stayed pending and **TTL-evicted (offset 0 lost)** — reproducing the exact gappy-bundle failure this PR targets.

**Fix:** record the `confirmed_data_roots` marker **unconditionally** at confirm time (so later-seeded chunks inherit it regardless of ordering), bounded by an age-based prune in the GC sweep (`CHUNK_INGEST_CONFIRMED_ROOT_RETENTION_SECONDS`, default 1 h).

### Run 2 (with the fix) — clean pass

| metric | result |
|---|---|
| chunks on disk | **1025 / 1025** (complete) |
| `relative_offset` 0 present | **yes** |
| confirmed / pending | **1025 / 0** |
| late-arrivals confirmed (cached after confirm) | **916** |
| `chunk_ingest_evicted_total{ttl}` | **0** |
| duration retained | **36 min** (~12× the 180 s TTL) |
| marker | present, table bounded (~517 rows, pruned) |
| unbundle | succeeded from local cache |
| data item | GraphQL-visible, `bundledIn` set |

The whole bundle was retained and unbundled with **zero network re-fetch**, sustained far past the TTL that would have evicted it on the old code.

**Caveat (honest):** Run 2's timing happened to land the first chunks ~4 s *before* confirm, so it exercised sibling inheritance + the marker with a few chunks present; the confirm-**before-any-chunk** path (Run 1's timing) is covered by the added unit test and by Run 1's demonstrated failure, rather than a second live run at that exact race.

vilenarios:
## Deterministic confirm-before-any-chunk validation

The earlier E2E runs happened to land the first chunks a few seconds *before* the confirm event, so they didn't stress the exact ordering that broke the pre-fix code. To force it deterministically, I bumped the bundler's `CHUNK_BROADCAST_GATE_MIN_CONFIRMATIONS` (1 → 3) so chunk seeding waits ~2 extra blocks after the gateway confirms.

**Captured the decisive window** (16 MiB / 65-chunk bundle):

- Gateway set the `confirmed_data_roots` marker with **`onDisk: 0` for ~73 seconds** — the data root was confirmed while **zero chunks** existed.
- When chunks finally seeded: **`confirmAtIntoWindow: -73s`** (confirm timestamp is 73 s *before* the first chunk's `cached_at`), **`lateArrivalsConfirmed: 65/65`** — every chunk arrived after confirm and was still confirmed.
- **65/65 complete, offset-0 present, 0 TTL evictions, held >3.7 min past the 180 s TTL.**

`lateArrivalsConfirmed = 100%` is the key: sibling inheritance is impossible here (no confirmed chunk existed at confirm time), so the retention can *only* come from the unconditional marker. This is precisely the timing that evicted all chunks on the pre-fix code.

### Test environment cleaned up
Gateway reverted to the released image + production TTL; bundler `MIN_CONFIRMATIONS` restored to 1. The one-row-per-confirmed-data-tx marker table is bounded by the age prune and observable via `chunk_ingest_confirmed_roots` / `chunk_ingest_confirmed_roots_pruned_total`.

vilenarios:
Both CodeRabbit findings addressed in 279b7926:

1. **Sibling-confirmation lookup O(n) → O(log n)** — added a partial index `chunk_placements_confirmed_sibling_idx ON chunk_placements(data_root) WHERE confirmed_at IS NOT NULL`. `EXPLAIN QUERY PLAN` now shows `SEARCH chunk_placements USING INDEX chunk_placements_confirmed_sibling_idx (data_root=?)` instead of a range scan. Put in a separate migration so it also applies on gateways where the marker-table migration was already recorded. (Kept the sibling fallback — it's not redundant with the marker: it extends retention for chunks that arrive after the marker's age-prune while a confirmed sibling is still on disk.)

2. **Unasserted `rec.prunedBefore`** — the GC eviction test now asserts `pruneConfirmedDataRoots` is called exactly once per sweep with a `now - retention` cutoff.

PR #818, opened 2026-07-08
fix(sqlite): stop getTxByOffset misses from scanning the offset index to the end

arielmelendez:
## Motivation

The SQLite slow-op log added in #817 captured `getTxByOffset` calls with
`queueWaitMs=0` but `serviceMs` of 23–110 **seconds** on a canary gateway, all
during retrieval-load episodes where peer socket pools pinned at their caps.
Re-running the captured offsets against a production-sized core DB showed every
slow call was a **miss** — an absolute weave offset in a coverage gap (format-1
tx or unpopulated region) — while hits consistently resolved in ~10ms.

Root cause: `selectStableTransactionOffsetById` keeps the span predicate
(`(offset - data_size) < @offset`) inside the index scan. On a hit the first
index row matches immediately. On a miss no row can ever match, so SQLite walks
`stable_transactions_offset_idx` from `@offset` to the end of the table — with a
table-row fetch per entry — before returning empty. Each such miss pins a core
read worker for tens of seconds and holds the requesting peer's socket open,
cascading into outbound socket-pool saturation.

## Fix

Weave data regions never overlap, so the **first** `format = 2, data_size > 0`
row with `offset >= @offset` is the only possible spanner — any other tx whose
region ends at or past `@offset` starts after that row's region ends. The
statement now fetches exactly that row in an inner `LIMIT 1` subquery and
applies the span check outside. A miss costs the same single index probe as a
hit.

**Partial-index caveat (guarded in code + test):** `stable_transactions_offset_idx`
is a partial index (`WHERE format = 2 AND data_size > 0`). SQLite only uses it
when the query's WHERE clause implies the index's WHERE clause, so both
predicates must remain inside the inner query verbatim — hoisting either one out
silently degrades to a full table scan. The SQL comment documents this and the
new test asserts the plan.

## Validation

Against a production-sized core DB (read-only, exact offsets captured by the
slow-op log):

| offset | before | after |
|---|---|---|
| hit (control) | ~10ms | ~10ms, identical row |
| miss (logged 44s / measured 41.8s cold) | 23–110s | **~10ms** |
| miss ×3 more | 23–27s logged | **~10ms** |
| beyond weave end | full remaining scan | **~10ms** |

`EXPLAIN QUERY PLAN` confirms a single
`SEARCH stable_transactions USING INDEX stable_transactions_offset_idx (offset>?)`.

## Tests

New regression test covering: hit, gap miss, format-1-spanned miss, beyond-end
miss, and an `EXPLAIN QUERY PLAN` assertion (statement loaded via `sql-loader`,
not duplicated) that the plan probes `stable_transactions_offset_idx` and never
scans `stable_transactions` — this specifically catches the partial-index
mismatch, which timing on a small test DB cannot. Full
`standalone-sqlite.test.ts` suite: 69/69 pass.

### DIFF
--- a/src/database/sql/core/offsets.sql
+++ b/src/database/sql/core/offsets.sql
@@ -1,12 +1,33 @@
 -- selectStableTransactionOffsetById
+-- Resolve an absolute weave offset to the stable format-2 transaction that
+-- spans it. `offset` is the tx's END offset, and tx data regions never overlap,
+-- so the FIRST format-2, data_size > 0 row (by offset ASC) with
+-- offset >= @offset is the only possible spanner: any other tx whose region
+-- ends at or past @offset starts after the spanner's region ends. The inner
+-- query fetches exactly that one row, and only the span predicate is applied
+-- outside. Keeping the span predicate inside the scan (the previous form) made
+-- a MISS -- an offset in a coverage gap (format-1 tx, unpopulated region) --
+-- walk the index from @offset to the end of the table looking for a match that
+-- cannot exist (tens of seconds on a large DB, pinning a core read worker);
+-- with this form a miss costs the same single index probe as a hit and returns
+-- empty.
+--
+-- IMPORTANT: stable_transactions_offset_idx is a PARTIAL index
+-- (WHERE format = 2 AND data_size > 0). Both predicates must stay inside the
+-- inner query verbatim -- SQLite only uses a partial index when the query's
+-- WHERE clause implies the index's WHERE clause; hoisting either predicate out
+-- silently degrades this to a full table scan.
 SELECT id, data_root, offset, data_size
-FROM stable_transactions
-WHERE offset >= @offset
-  AND (offset - data_size) < @offset
-  AND format = 2
-  AND data_size > 0
-ORDER BY offset ASC
-LIMIT 1;
+FROM (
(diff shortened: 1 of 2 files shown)


arielmelendez:
@coderabbitai Addressed the nitpick in f53145fd — switched to `fileURLToPath(new URL(...))` for the SQL directory path (with the `node:url` import). On the second point: the suite passing confirms the `.sql` files resolve at test runtime (tests execute against `src/` via the TS loader, not `dist/`). Note `standalone-sqlite.ts` itself uses the same `.pathname` pattern when loading these statements; left that as-is since changing production code is out of scope here.

arielmelendez:
**Disjointness assumption validated against production data.**

The rewrite assumes tx data regions in `stable_transactions` never overlap (first candidate row = only possible spanner). Ran a full read-only audit on a canary gateway's production core DB: for all **70,127,446** rows in the partial-index domain (`format = 2 AND data_size > 0`), ordered by end offset, zero rows have `(offset - data_size) < LAG(offset)` — i.e. **0 overlap violations**. Any overlap between any two regions would necessarily surface as a consecutive-pair violation in this ordering, so the check is complete, not sampled.

On this dataset the new statement is therefore provably input-equivalent to the old one (identical hits and misses for every possible offset) — minus the pathological miss-scan cost.

PR #829, opened 2026-07-24
perf(root-tx): local-first offset resolution in RootParentDataSource (PE-9134)

arielmelendez:
**Stacked on #828** (`PE-9134-roottx-accept-predicate`) → #827. Base retargets down the stack as each lands. Part 2 of the follow-up — the first change that actually *uses* the accept predicate.

## What

`RootParentDataSource` legacy traversal now resolves the root TX **local-first**:

1. `getRootTx(id, { accept: r => r.rootTxId != null })` — short-circuits on the first source carrying a rootTxId (db/cdb), skipping remote sources.
2. Resolve offsets from the bundle header via the existing `getDataItemOffset` linear scan — bytes the source reads to serve the item anyway.
3. **Only if that local scan misses** (item nested beyond the root bundle's direct children) do we issue a full actionable `getRootTx(id)` (which may consult GraphQL) and retry with its path (path-guided nav) or direct offsets.

## Why

Validated on a canary: `l1_root` short-circuits at db carry most of the win, but the frozen historical CDB population is bare `{r}` (rootTxId only, no path/offsets, un-re-exportable), so those hits still fell through to GraphQL for a path shortcut. When GraphQL is slow, every one of those eats a multi-second round-trip for an item the CDB already located. This resolves them from the local bundle header instead.

## Safety — today's behavior is the floor

- **Shallow items** (direct children of the root bundle): resolve with **zero remote lookup**. Win.
- **Nested items**: identical to today (the on-miss fallback issues the same GraphQL path lookup), plus one cheap local scan. Cannot do worse.
- The not-found path still consults the full chain (incl. GraphQL) before erroring — no coverage regression.

The only way this regresses is a bug in the on-miss fallback, which the new metric + canary error watch catch quickly.

## Observability

`root_tx_local_resolve_total{outcome}` with `local` (resolved by local scan, GraphQL avoided) / `remote_fallback` (local miss, recovered a path or offsets) / `unresolved`. The `local` share is the direct measure of GraphQL avoidance.

## Tests

New: local-scan short-circuit (asserts a single, rootId-accepting lookup and no remote fallback) and the on-miss path fallback (asserts the second default lookup + path-guided retry). Updated the existing not-found test for the extra fallback lookup. 38/38 pass.

## Rollout

Canary on one gateway, watch `root_tx_local_resolve_total` (expect a healthy `local` share), `caller_accept` exit reason, GraphQL probe rate, and error rate. `tx-metadata-resolver` gets the same treatment in a follow-up if this looks good.

### DIFF
--- a/src/data/root-parent-data-source.test.ts
+++ b/src/data/root-parent-data-source.test.ts
@@ -17,6 +17,17 @@ import {
   DataItemRootIndex,
 } from '../types.js';
 import { createTestLogger } from '../../test/test-logger.js';
+import * as metrics from '../metrics.js';
+
+// Reads the current value of root_tx_local_resolve_total for a given outcome
+// label so tests can assert the increment (the counter is process-global, so
+// assert deltas rather than absolute values).
+async function readLocalResolve(outcome: string): Promise<number> {
+  const metric = await metrics.rootTxLocalResolveTotal.get();
+  return (
+    metric.values.find((v: any) => v.labels.outcome === outcome)?.value ?? 0
+  );
+}
 
 describe('RootParentDataSource', () => {
   let log: winston.Logger;
@@ -2362,7 +2373,9 @@ describe('RootParentDataSource', () => {
         cached: false,
       }));
 
+      const before = await readLocalResolve('local');
       await rootParentDataSource.getData({ id: dataItemId });
+      assert.strictEqual((await readLocalResolve('local')) - before, 1);
 
       // Exactly one root-tx lookup: the local scan succeeded, so no remote
       // fallback lookup was issued.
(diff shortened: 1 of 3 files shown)


arielmelendez:
@coderabbitai review

arielmelendez:
@coderabbitai review

arielmelendez:
Addressed all four review nitpicks in 00f09185:

1. **Dedicated span for the remote fallback** — the fallback lookup is now in `resolveRemoteFallbackOffset`, which opens its own `RootParentDataSource.remoteFallbackOffset` child span, so the slow-path latency is independently visible in traces.
2. **Extract the remote-fallback into a helper** — done (`resolveRemoteFallbackOffset`), removing the extra nesting from `getData`.
3. **Metric doc in the `help` field** — folded the `local`/`remote_fallback`/`unresolved` descriptions into `help` and dropped the separate comment, matching the file convention.
4. **Test coverage** — added tests for the `linear_then_offsets_fallback` (direct-offsets) branch and the `unresolved` outcome, and all three local-first tests now assert the `root_tx_local_resolve_total` label increment via a registry snapshot delta.

40/40 tests pass; eslint/prettier clean; tsc unchanged.

PR #832, opened 2026-08-04
feat(manifest): faster + more correct manifest path resolution (cache, index precedence, id validation)

vilenarios:
Draft — PR 1 of 2 in the manifest-resolution work (see follow-up for the persistent `resolveFromIndex` index). No schema changes here.

## What & why

Every ArNS site is an Arweave **manifest** (a map of URL paths → data ids), and each asset on a page is a *separate* path resolution against the same manifest. Today the gateway streams and re-parses the entire manifest body **on every resolution**. This PR removes that redundant work and fixes two correctness issues, all within the streaming resolver — no DB/migration.

## Changes

**`perf`: cache resolved manifest paths (1c).**
A manifest tx is immutable, so a resolved `(manifest id, path) → data id` mapping is valid forever. `StreamingManifestPathResolver` now keeps a bounded LRU: `resolveFromData` populates it; `resolveFromIndex` serves hits with `complete: true`, so the caller (`sendManifestResponse`) skips fetching **and** re-parsing the body. Positive and negative resolutions are both cached. Size via `MANIFEST_RESOLUTION_CACHE_SIZE` (default 5000).

**`fix`: deterministic v0.2.0 index precedence (1a).**
Index resolution is deferred until the `index` object closes, so `index.id` deterministically wins over `index.path` regardless of JSON key order. Previously a manifest whose `paths` preceded `index` (with `path` before `id`) resolved to the id `index.path` mapped to — meaning the served index could differ from other gateways (a cross-gateway consistency bug). Depth guard excludes a `paths` entry keyed `"index"`.

**`fix`: reject malformed data ids (1b).**
Manifest entries whose id isn't 43-char base64url are rejected before a data-retrieval attempt.

## Before/after (real manifests, 200 repeat resolutions, body in memory → isolates parse cost)

| manifest | before | after | speedup | parse cost/call today |
|---|--:|--:|--:|--:|
| small (100 paths, 9 KB) | 129 ms | 5.8 ms | 22× | 0.65 ms |
| medium (269 paths, 28 KB) | 285 ms | 3.8 ms | 76× | 1.4 ms |
| large (7,784 paths, 500 KB) | 11,363 ms | 59 ms | **192×** | **57 ms** |

The large real manifest costs ~57 ms of CPU parse **per resolution** today — a 30-asset homepage burns ~1.7 s of pure parse CPU per view, eliminated on cache hits. When the body is *not* locally cached, the current path additionally pays a full upstream fetch per resolution (not measured here) — the follow-up persistent index removes that too.

## Tests
- `src/lib/encoding.test.ts`: +4 (deterministic index-id precedence regression; malformed-id rejection). 29/29 pass.
- `src/resolution/streaming-manifest-path-resolver.test.ts` (new): cache hit/miss, negative caching, trailing-slash key normalization, per-manifest keying. 5/5 pass.
- `lint:check` clean.

## Docs
`MANIFEST_RESOLUTION_CACHE_SIZE` added to `docs/envs.md` and `docker-compose.yaml`.

## Not in this PR
- `1e` observability metrics (small follow-up).
- `1d` persistent `manifest_index_id`/`manifest_fallback_id` index implementing `resolveFromIndex` from a dedicated table (separate PR — carries the migration).


vilenarios:
Superseded by #833. That PR now contains these commits (1a deterministic index precedence, 1b id validation, 1c resolution cache) **plus** the persistent index (1d), as a single self-contained PR against `develop` so CodeRabbit reviews the full change. No code is lost — #833 includes everything here.

PR #833, opened 2026-08-04
feat(manifest): faster, correct, and resilient manifest path resolution

vilenarios:
Supersedes #832 — this PR now contains those commits (1a–1c) **plus** the persistent index (1d) as a single, self-contained change against `develop`. No schema entanglement beyond one dedicated table.

Every ArNS site is an Arweave **manifest** (URL paths → data ids), and each asset on a page is a separate path resolution against the same manifest. This PR fixes two correctness bugs, removes redundant re-parsing, and makes resolution survive restarts and upstream outages.

## Changes

**`fix` — deterministic v0.2.0 index precedence (1a).** Index resolution is deferred until the `index` object closes, so `index.id` wins over `index.path` regardless of JSON key order. Previously a manifest whose `paths` preceded `index` (with `path` before `id`) resolved to the wrong id — the served index could differ from other gateways. (Bug reproduced, then fixed; regression test added. Depth guard excludes a `paths` entry keyed `"index"`.)

**`fix` — reject malformed ids (1b).** Manifest entries whose id isn't 43-char base64url are rejected before a data-retrieval attempt.

**`perf` — in-memory resolution cache (1c).** A manifest tx is immutable, so a resolved `(manifest id, path) → data id` mapping is valid forever. `StreamingManifestPathResolver` keeps a bounded LRU; repeat resolutions skip fetch + parse. Size via `MANIFEST_RESOLUTION_CACHE_SIZE` (default 5000).

**`feat` — persistent index (1d).** Implements the previously-stubbed `resolveFromIndex` against a durable store: a new `manifest_resolutions` table (`data.db`, dedicated — no `stable_*`/auto-verify/flush/ClickHouse entanglement) keyed by manifest tx id → `index_id`/`fallback_id`, with `getManifestResolution`/`saveManifestResolution` on the data worker pool behind a `ManifestResolutionStore` interface. The resolver consults the store for the **root only** (index/fallback — mirrors legacy `handleManifestFromDdb`) and lazily persists on resolve. Fire-and-forget; store failures never break serving. Additive migration with a reversing down migration.

## Before/after (real manifests from a live gateway)

Repeat resolutions, body in memory (isolates parse cost):

| manifest | before | after (cache) | speedup |
|---|--:|--:|--:|
| small (100 paths, 9 KB) | 129 ms | 5.8 ms | 22× |
| medium (269 paths, 28 KB) | 285 ms | 3.8 ms | 76× |
| large (7,784 paths, 500 KB) | 11,363 ms | 59 ms | **192×** |

Cold cache / after restart (persistent index), large manifest:

| scenario | before | after |
|---|--:|--:|
| root serve, cold/after-restart | ~57 ms re-parse | **~4.4 µs** store lookup (~12,800×) |
| root serve, body unreachable upstream | **404** | **served** from local index |

## Tests
- `encoding.test.ts` +4 (index-id precedence regression; malformed-id rejection) — 29/29.
- `streaming-manifest-path-resolver.test.ts` (new, 12): cache hit/miss/negative, trailing-slash keying, store hit for index/fallback, sub-path skip, lazy persist, store-failure fallthrough, store-optional.
- `manifest-resolutions.test.ts` (new, 4): real-SQLite round-trip incl. COALESCE upsert, missing-manifest.
- `standalone-sqlite.test.ts` 69/69 (no worker regression). Lint clean; zero typecheck errors in changed files.

## Migration
`migrations/2026.08.04T12.00.00.data.init-manifest-resolutions.sql` (+ `down/`). `test/data-schema.sql` regenerated (only the new table).

## Behavioral notes for review
- Root manifest requests now do a fast SQLite PK lookup before body fetch (guarded; non-manifest paths untouched).
- A fire-and-forget write persists the root resolution on the request path (bounded ~once per manifest, errors swallowed).

## Not in this PR
`1e` observability metrics (small follow-up). Eager population at unbundle time deliberately deferred (lazy-on-request covers L1 + bundled uniformly).

New env var `MANIFEST_RESOLUTION_CACHE_SIZE` documented in `docs/envs.md` + `docker-compose.yaml`.


vilenarios:
@coderabbitai full review

vilenarios:
Thanks @coderabbitai — addressed in d964bd58:

**1. `queueWork` orphaned `.finally()` promise** — ✅ fixed. `queueWork` now returns the `.finally()`-chained promise instead of the bare `ret`, so a rejection is delivered to the caller's handler rather than surfacing as an unhandledRejection. `.finally()` passes settlement through unchanged, so behavior is preserved; the full DB suite (standalone-sqlite + chunk-placements + manifest-resolutions) stays green at 79/79. Good catch — this was latent but newly reachable via the store's fire-and-forget write.

**2. Fallback resolution ordering** — verified as already-correct, and added a proof. `parseManifestStream` emits the `fallback` event only at stream `end` (after every `index`/`path` event), so a fallback that appears *first* in the JSON cannot preempt them. Added `example_manifest_v0.2.0_fallback_first.json` + tests asserting the root resolves to the index (not fallback), a matching sub-path resolves to that path (not fallback), and the fallback is used only on a miss. No logic change needed.

**3. `getManifestResolution` switch-case braces** — intentionally skipped for consistency. This switch uses the unwrapped `case '…': const x = worker.…(args[0]); parentPort?.postMessage(x); break;` idiom for ~20 existing cases (`getDataAttributes`, `getDataItemAttributes`, …), and repo ESLint passes. Per the project's CLAUDE.md ("write code that reads like the surrounding code"), I've kept the new case consistent rather than introducing a lone braced case.

PR #835, opened 2026-08-05
feat(manifest): observability metrics for path resolution

vilenarios:
Adds the observability layer (1e) for manifest path resolution. Independent of the resolution PR (#833) — touches only `metrics.ts` and the data handler, no overlap — and is meaningful now (data-path metrics) and richer once #833 lands (index-path metrics start firing).

## Metrics
- **`manifest_resolutions_total{source,resolution_type}`** — resolutions by `source` (`index` = served from the persistent/cached index without parsing the body, vs `data` = on-demand body parse) and `resolution_type` (`path`/`index`/`fallback`/`unresolved`). The **index-vs-data ratio** is the effectiveness signal for the resolution index + cache.
- **`manifest_unresolved_root_total`** — root/index requests that resolved to nothing (a malformed-manifest signal that today only surfaces as a user 404).
- **`manifest_resolution_duration_seconds{source}`** — resolution latency by source (bounded lookup vs body parse).

## Placement
Instruments the two existing resolution sites in `src/routes/data/handlers.ts`. `index` is counted only when the index actually determined the outcome (`complete`); a miss falls through and is counted at the data-parse site, so each resolution is counted once.

## Tests
`manifest-resolution-metrics.test.ts` (new, 4): verifies label routing (source × resolution_type), duration observation, and that the unresolved-root counter fires for an unresolved **root** but not for an unresolved **sub-path**. Handler suite unaffected (96/96); lint clean; zero new type errors.

## Notes
No new env var and no metrics doc exists in `docs/`, so nothing to sync there.

### DIFF
--- a/src/metrics.ts
+++ b/src/metrics.ts
@@ -1806,3 +1806,35 @@ export const optimisticTxVerificationBlockedCounter = new promClient.Counter({
   name: 'optimistic_tx_verification_blocked_total',
   help: 'Count of verification attempts withheld because the root tx is not yet mined (optimistic / NULL height)',
 });
+
+//
+// Manifest path resolution
+//
+
+// Count of manifest path resolutions, split by `source` — `index` (served
+// from the persistent/cached index without parsing the body) vs `data`
+// (on-demand body parse) — and by how the path resolved (`resolution_type`:
+// path / index / fallback / unresolved). The index-vs-data ratio is the
+// effectiveness signal for the resolution index and cache.
+export const manifestResolutionsTotal = new promClient.Counter({
+  name: 'manifest_resolutions_total',
+  help: 'Manifest path resolutions, by source (index vs on-demand data parse) and resolution type',
+  labelNames: ['source', 'resolution_type'] as const,
+});
+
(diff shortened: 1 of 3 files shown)


vilenarios:
@coderabbitai review

vilenarios:
Thanks @coderabbitai — addressed in fac809f0:

**1. `metrics.ts` comments** — converted the metric-definition comments from `//` to `/** */` block form, matching the existing convention in this file (e.g. the `optimisticTx*` counters carry multi-line `/** */` rationale blocks). I kept the operational context — the index-vs-data effectiveness signal and the malformed-manifest signal — since that's guidance not expressible through the one-line `help` fields, and it's consistent with how neighboring metrics are documented here.

**2. `recordManifestResolutionMetrics`** — replaced the line-comment block with TSDoc documenting `source`, `resolution`, `manifestPath`, and `durationMs`, including that an unresolved root increments `manifest_unresolved_root_total`.

**3. Test helpers** — added TSDoc to `counterValue`, `histogramCount`, and `resolution` describing the prom-client sample shape, label matching, the 0 fallback, and the default resolution.

Lint clean, metrics tests 4/4.

PR #857, opened 2026-08-12
fix(compose): forward the observer's cranker and cleanup settings

vilenarios:
The observer reads ten settings that `docker-compose.yaml` never passed into its container, so setting any of them in `.env` silently did nothing:

```
CRANK_BATCH_SIZE            CLEANUP_BATCH_SIZE
CRANK_CLOSE_EPOCHS          CLEANUP_FAILURE_THRESHOLD
CRANK_EPOCH_RETENTION       MAX_CLEANUP_TXS_PER_CYCLE
CRANK_WARN_BALANCE_SOL      ALT_RECLAIM_SCAN_LIMIT
CRANK_CRITICAL_BALANCE_SOL  OBSERVED_GATEWAY_HOSTS
```

Same gap #842 fixed for `CRANK_POLL_INTERVAL_MS` / `CLEANUP_MIN_INTERVAL_MS` and #846 fixed for `CLEANUP_TO_RETURNED_TXS_PER_CYCLE` — the rest of the family was never wired.

## What prompted it

`CRANK_EPOCH_RETENTION` governs how many epochs survive before `close_epoch` reclaims their rent, and closing an epoch also closes its Observation PDAs — which is what removes per-observer report detail from chain.

Measured on mainnet: only the **two most recent epochs** still had Observation accounts (13 for epoch 510, 9 for 511, nothing older), while the Epoch PDAs for 506–509 survived holding only aggregate `observationsSubmitted` counts. An operator wanting more history had no way to ask for it.

`ALT_RECLAIM_SCAN_LIMIT` matters for the same reason from the other direction: reclaiming a prescribe Address Lookup Table removes the observer index for its epoch, so `getEpoch` can no longer enumerate that epoch's reports even while the underlying accounts are alive.

Worth noting for anyone reading this later: `close_observation` and `close_epoch` are **permissionless**, so the effective retention across the network is set by the most aggressive cranker running, not by any single operator. This PR makes the setting reachable; it doesn't make it authoritative.

## Regression check

This one had a real failure mode worth verifying rather than assuming. `parsePositiveIntEnv` does `Number.parseInt(raw)` and **throws** on a non-positive integer — so if `${VAR:-}` rendered an empty string that reached it, the observer would fail to start on every deployment that doesn't set these.

Verified against the running observer image (`15e285b0`):

- `env.varOrDefault` returns the supplied default for both **empty** and **whitespace-only** values, so a blank never reaches the numeric parsers.
- Loading `dist/config.js` with all ten set to `''` yields every documented default unchanged:

```
CRANK_BATCH_SIZE            '' -> 15      CLEANUP_BATCH_SIZE         '' -> 15
CRANK_CLOSE_EPOCHS          '' -> true    CLEANUP_FAILURE_THRESHOLD  '' -> 30
CRANK_EPOCH_RETENTION       '' -> 7       MAX_CLEANUP_TXS_PER_CYCLE  '' -> 50
CRANK_WARN_BALANCE_SOL      '' -> 0.3     ALT_RECLAIM_SCAN_LIMIT     '' -> 200
CRANK_CRITICAL_BALANCE_SOL  '' -> 0.1
```

Also confirmed `docker compose config` renders the values through when set (`CRANK_EPOCH_RETENTION: "14"`, `ALT_RECLAIM_SCAN_LIMIT: "0"`) and empty when unset.

## Scope

Seven observer settings remain unforwarded — `PORT` (intentional; compose maps `OBSERVER_PORT`), `ARNS_NAMES`, `ALWAYS_SAVE_REPORTS`, `REFERENCE_GATEWAY_HOST`, `BLOCK_OFFSET_MAPPING_ENABLED`, `TX_PATH_PARSING_ENABLED`, `ENABLE_OPENAPI_VALIDATION`. Those are observer-internal/dev knobs rather than operational cranker settings, so I left them out to keep this coherent. Happy to add them if preferred.

Documented in `docs/envs.md` per the compose/envs.md sync rule.


vilenarios:
Handled the CodeRabbit finding — it was correct. Verified against the `15e285b0` observer image:

```
'host-a, host-b'  ->  ["host-a"," host-b"]   <- leading space, matches nothing
'host-a,host-b'   ->  ["host-a","host-b"]
```

Worth noting it fails *silently*: the malformed entry just never matches a gateway, so the observation set narrows with no error.

Addressed both halves of the suggestion rather than picking one:

1. **Here (62afa1e5):** documented the no-space constraint, since the parsing lives in ar-io-observer and can't be fixed from this repo.
2. **Root cause — ar-io/ar-io-observer#123:** adds the trim. `REFERENCE_GATEWAY_HOSTS`, defined directly above `OBSERVED_GATEWAY_HOSTS` in the same file, already trims; the other two list vars diverged from it. `ARNS_NAMES` had the identical bug and is fixed too. 375 tests passing, 6 new covering the reported input.

Once #123 ships, the doc note here can be dropped.

Also flagging why this surfaced now: `OBSERVED_GATEWAY_HOSTS` was never forwarded into the observer container before this PR, so the footgun was unreachable via compose and is about to become reachable.

PR #866, opened 2026-08-20
docs(cache): ADR for indexed eviction of the chunk data cache

vilenarios:
## Summary

ADR proposing indexed eviction for the chunk data cache, plus corroborating measurements from the second gateway.

The chunk cache is reclaimed only by `FsCleanupWorker`, the filesystem-walk reclaimer. Its sibling — the contiguous data cache — has had an indexed evictor since PE-9131, and `docs/cache-cleanup.md` already recommends the index for large caches, stating the walk "cannot keep up" on them. The chunk cache is the larger of the two by object count and has no evictor.

## Measured on both gateways, independently

| | gw1 | gw2 |
|---|---|---|
| time spent walking per cycle | 93.9s of 94.9s (99%) | 67.7s median gap, 250ms pause floor |
| batch multiplier | at 4× cap | at 4× cap |
| delete throughput | ~60 GB/hr | 8.2 GB/hr |
| device utilisation | 3.4% | 1.84% |

Discovery — not deletion, not I/O — is the constraint on both. Every other lever is exhausted: batch size at its cap, pause effectively zero, walk concurrency at the safe ceiling.

**PR #847 (watermarks) does not resolve this**, and this ADR treats it as the stopgap it is. On gw2, with #847 deployed, the volume still reached 99% twice in 24 hours and required two manual `find -delete` sweeps. Watermarks scale batch size; they cannot make discovery faster.

## What the ADR decides

Port `ContiguousDataCacheEvictor` / `ContiguousDataCacheReconciler` to the chunk cache, **keyed on dataRoot rather than per-chunk** (~37,581 rows — smaller than the live contiguous index — versus ~4.5M), with two mandatory departures:

1. **The evictor must honour an age floor.** The contiguous evictor deliberately ignores `AGGRESSIVE_MIN_AGE_SECONDS`. That is safe for contiguous data and **not** safe for chunks: the floor is what prevents evicting a chunk before its placement confirms. Porting the contiguous semantics verbatim would introduce a data-propagation bug. This is the most important review point in the document.
2. **Eviction must be size-aware, with a hybrid tail.** The object distribution is extreme — p50 is 2 chunks, max is 11,218, and the top 1% of dataRoots hold 55.7% of all chunks. Naive coldest-first LRU would repeatedly select 0.5 MiB entries and reclaim nothing, while a single hot chunk could pin a 2.69 GiB entry resident.

The reuse profile drives retention and shows the volume is not undersized: **97.8% of reuse occurs within 15 minutes, 64.5% of chunks are never read again**, and the volume holds ~9 hours of history against a ~0.5 hour reuse need — roughly 18× more capacity than the workload requires. The failure is reclaim throughput, not capacity.

## Notes for reviewers

- `chunk_placements` (optimistic ingest path) already indexes **20.4%** of the cache on gw2 — 876,850 rows against ~4.29M files, in a separate `chunks.db` with its own WAL. Not a vehicle for eviction (per-chunk, which this ADR rejects), but prior art for the write hook and a precedent for keeping this bookkeeping off `data.db`.
- The ADR records a methodology caveat: inferring re-fetch rates by differencing counters gave an implausible result on gw2 because the delete term missed `FsChunkDataStore.del()`. The `atime − mtime` sampling used throughout is the sound method.
- Index/FS drift is a live concern, not theoretical — gw2's contiguous index was empty and freed 0 bytes per sweep; gw1 logged "Cache index drained but still over pressure" 51× per 2h. The reconciler is not optional at chunk cardinality.

https://claude.ai/code/session_01WndVj5cprBtfa5d2qqnacp

### DIFF
--- a/docs/INDEX.md
+++ b/docs/INDEX.md
@@ -85,6 +85,7 @@ Technical details about Arweave data structures.
 | [002 - ArNS Cache Timing](madr/002-arns-cache-timing.md) | ArNS resolution caching strategy |
 | [003 - ArNS Undername Limits](madr/003-arns-undername-limits.md) | Undername resolution limits |
 | [004 - Optimistic L1 Transaction Indexing](madr/004-optimistic-l1-tx-indexing.md) | Index a signed L1 tx before it mines + the never-serve-as-permanent guard |
+| [005 - Chunk Data Cache Indexed Eviction](madr/005-chunk-data-cache-indexed-eviction.md) | Port the contiguous cache's indexed evictor to the chunk cache (proposed) |
 
 ## Testing
 
--- a/docs/cache-cleanup.md
+++ b/docs/cache-cleanup.md
@@ -116,3 +116,12 @@ Recommended large-cache setup:
   worker doesn't also run (or set it if you want it as an occasional reconciler).
 
 See [Environment Variables](envs.md) for the full list of related settings.
+
+## Chunk data cache
+
+This document covers the **contiguous** data cache only. The chunk data cache
+(`data/chunks`) has no index evictor — it is reclaimed solely by
+`FsCleanupWorker`, which on a cache of several million chunks is walk-bound for
+the reasons described above. See
+[madr/005-chunk-data-cache-indexed-eviction.md](madr/005-chunk-data-cache-indexed-eviction.md)
+for measurements and a proposal to port this pattern there.
--- a/docs/madr/005-chunk-data-cache-indexed-eviction.md
+++ b/docs/madr/005-chunk-data-cache-indexed-eviction.md
@@ -1,0 +1,264 @@
+# Indexed Eviction for the Chunk Data Cache
+
+- Status: proposed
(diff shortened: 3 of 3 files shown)


vilenarios:
Superseded by #870, which implements this ADR and carries an amended copy of it at `docs/madr/005-chunk-data-cache-indexed-eviction.md`.

The ADR moved from `proposed` to `accepted (implemented)`, with the design record updated where review disagreed with the proposal — most importantly the schema: the age-floor column is `last_write` (max write time, advanced on every upsert), **not** `cached_at`. Leaving it immutable-on-upsert, as the contiguous index does, is a correctness bug at data-root granularity, since eviction is all-or-nothing per directory.

Two further defects of the same shape were found and are recorded there as resolved (floor checked only at selection time; backfill emitting partial rows on abort), along with the production measurements and the resolution of the absolute-offset symlink question.

Closing in favour of #870 so the ADR and its implementation stay in one place.

PR #867, opened 2026-08-20
fix(chunks): retry a chunk write when its data-root directory disappears

vilenarios:
## Problem

`FsChunkDataStore.set()` creates the data-root directory, then writes the chunk into it:

```js
await fs.promises.mkdir(chunkDataRootDir, { recursive: true });
const chunkPath = this.chunkDataRootPath(dataRoot, relativeOffset);
await fs.promises.writeFile(chunkPath, chunkData.chunk);
```

If the directory is removed between those two calls, `writeFile` fails `ENOENT`, the outer `catch` swallows it (the index is best-effort, so failures are logged and not rethrown), and **the chunk is silently dropped** — the caller is told nothing and believes it was cached.

## Why now

This is **latent, not live**: nothing reaps empty data-root directories today, which is why 67% of them are empty on a production gateway.

It stops being latent the moment eviction works at data-root granularity — which is exactly what [ADR 005](https://github.com/ar-io/ar-io-node/blob/docs/chunk-cache-indexed-eviction/docs/madr/005-chunk-data-cache-indexed-eviction.md) proposes. That ADR flags this as a prerequisite:

> *"note `FsChunkDataStore.set()` does `mkdir` then `writeFile` with **no ENOENT retry** — removing a directory between those two calls silently drops the chunk. Any `rmdir` work must add that retry first."*

Landing it separately means the eviction work doesn't have to carry a silent-data-loss fix alongside a design change, and the fix stands on its own regardless of whether directory reaping is ever built.

## Change

Recreate the directory and retry the write once on `ENOENT`. Anything else rethrows, so unrelated write failures still reach the existing error path unchanged. No behaviour change on the happy path — one extra `try` around a call that already succeeds.

## Testing

The new test simulates the directory vanishing mid-`set()` and asserts the chunk is still readable afterwards. Confirmed to fail without the fix:

```
without: not ok 1 - should retry the write when the data-root directory disappears mid-set
         19 pass / 1 fail
with:    20 pass / 0 fail
```

Type-clean; no new lint.

https://claude.ai/code/session_01WndVj5cprBtfa5d2qqnacp

### DIFF
--- a/src/store/fs-chunk-data-store.test.ts
+++ b/src/store/fs-chunk-data-store.test.ts
@@ -35,6 +35,45 @@ describe('FsChunkDataStore', () => {
     rmSync(tempDir, { recursive: true, force: true });
   });
 
+  describe('write durability', () => {
+    it('should retry the write when the data-root directory disappears mid-set', async () => {
+      const dataRoot = 'wRq6f05oRupfTW_M5dcYBtwK5P8rSNYu20vC6D_o-M4';
+      const relativeOffset = 0;
+      const chunkData: ChunkData = {
+        chunk: Buffer.from('survives a concurrent rmdir'),
+        hash: crypto
+          .createHash('sha256')
+          .update('survives a concurrent rmdir')
+          .digest(),
+      };
+      const fsp = (await import('node:fs')).promises;
+
+      // Simulate the directory being reaped between mkdir and writeFile, which
+      // is what data-root-granularity eviction will do. Without a retry the
+      // ENOENT is swallowed and the chunk is silently lost.
+      const realWriteFile = fsp.writeFile.bind(fsp);
+      let firstAttempt = true;
+      (fsp as any).writeFile = async (p: any, d: any) => {
+        if (firstAttempt && String(p).includes('by-dataroot')) {
+          firstAttempt = false;
+          const err: any = new Error('ENOENT: no such file or directory');
+          err.code = 'ENOENT';
+          throw err;
+        }
+        return realWriteFile(p, d);
+      };
+      try {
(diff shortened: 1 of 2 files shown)


vilenarios:
### Note on the review comments above — they are not about this PR

Apologies for the noise: the 11 inline CodeRabbit comments on this PR are **stale and unrelated to its diff**. They landed while a stack of chunk-cache eviction commits was briefly pushed onto this branch by mistake; the branch has since been reset to its original single commit. GitHub did not mark them outdated, so they still render against files that are not in this PR.

Nine of the eleven point at files this PR does not touch — `src/config.ts`, `src/workers/chunk-data-cache-reconciler.ts`, `docker-compose.yaml`, `docs/cache-cleanup.md`, `docs/glossary.md`, `docs/madr/005-*.md`, `src/database/standalone-sqlite.chunk-data-cache.test.ts`. Those findings were real and have all been addressed in #870, which is where that work now lives.

**This PR is unchanged from its original scope:**

```
src/store/fs-chunk-data-store.ts       +14/-1
src/store/fs-chunk-data-store.test.ts  +39/-0
```

Verified just now: 1 commit, CI green on ubuntu and macOS, `behind_by: 0` against `develop` (no rebase needed), no merge conflicts. The included test was checked against a deliberately reverted fix and fails without it, so it is not vacuous.

One thing worth flagging for merge ordering: **#870 is stacked on this branch and depends on this fix landing first.** Data-root-granular eviction removes directories, which races `set()`'s `mkdir` -> `writeFile`; this retry is what closes that window. An image that ships the evictor without this retry will silently drop chunks. #870 retargets to `develop` automatically once this merges.

PR #869, opened 2026-08-20
fix(x402): authenticate the CDP facilitator with the API key id

vilenarios:
## The bug

Every x402 payment on mainnet fails verification with `payment_verification_failed`.

`createFacilitatorConfig(apiKeyId, apiKeySecret)` expects a CDP **API key id**, but both
call sites pass `X_402_CDP_CLIENT_KEY` — the public **Client API key**, which `config.ts`
itself labels *"Paywall customization (optional)"*. CDP responds **401 Unauthorized**.

Meanwhile `CDP_API_KEY_ID`, the credential that actually authenticates, is passed into the
container by `docker-compose.yaml` and **read nowhere in `src/`**.

## Why it is hard to spot

The 401 never reaches the operator in a recognisable form. It surfaces to the client as a
generic `payment_verification_failed`, and the real reason appears only in the gateway's
own logs:

```
error: Payment verification error {"class":"X402UsdcProcessor",
  "error":"Failed to verify payment: Unauthorized"}
warn:  Payment verification failed {"reason":"Payment verification failed"}
```

Nothing in the response distinguishes a wrong credential from a malformed payment. An
operator sees x402 enabled, `/ar-io/info` advertising it, 402s being issued correctly, and
simply no revenue.

There is a second trap behind it: when CDP credentials are absent,
`createFacilitatorConfigFromCredentials` silently falls back to
`X_402_USDC_FACILITATOR_URL`. The commonly configured `https://facilitator.x402.rs`
supports 19 networks, **none of them Base mainnet** — the only `8453` in its `/supported`
response is a substring of `84532` (base-sepolia). So a misconfigured credential degrades
to an endpoint that also cannot settle, with no signal either way.

## Evidence

Both pairings probed against the live facilitator with identical requirements and a
deliberately bogus payment payload:

| apiKeyId passed | facilitator response |
|---|---|
| `X_402_CDP_CLIENT_KEY` (as shipped) | **401 Unauthorized** — credential rejected |
| `CDP_API_KEY_ID` (this PR) | **400 Bad Request** — authenticated, then evaluated the payload |

401 is "who are you"; 400 is "your payment is malformed". Only the second means auth
succeeded.

## The fix

Adds `X_402_CDP_API_KEY_ID` bound to `CDP_API_KEY_ID`, points both call sites at it, and
renames the `facilitator-utils` parameters (`cdpClientKey` → `cdpApiKeyId`) so the contract
cannot be misread again. `X_402_CDP_CLIENT_KEY` keeps its real job — the paywall UI.

## Verified end to end on Base mainnet

A 36 MB video paid for and served, after the fix:

```
402 -> sign -> retry -> 200
36,399,611 bytes of video/mp4
settlement 0x05b364bd5a43bcd31e1e7469e6d31acb36910d93f1c616951e484af0e732da59
block 50,218,636 · success · gas paid by the facilitator (EIP-3009 is gasless for the payer)
payer -0.004 USDC · operator +0.004 USDC
```

Before the fix the identical request returned 402 with `payment_verification_failed` and
0 USDC moved.

## Tooling

Two harnesses, because this path had no coverage on the wire and unit tests cannot reach it.

**`tools/x402-probe`** — reaches the 402 without moving a gigabyte of data. A 402 is only
emitted once a client's bucket is exhausted *and* no payment is attached, and an operator's
own egress IP is usually allowlisted, which skips the check entirely. The probe talks to
envoy directly (the public edge rewrites `X-Forwarded-For`, so a synthetic client IP only
survives there) and seeds a reserved TEST-NET-3 address's bucket to empty in redis. It
checks **both** shapes the processor branches on — JSON requirements for an agent, the HTML
paywall for a browser — and asserts an allowlisted client is still served, which makes it a
regression test for allowlist edits.

**`tools/x402-pay-test`** — drives a real wallet through 402 → sign → retry → 200, refuses
to run when payer equals `payTo` (a self-transfer proves nothing), enforces a `MAX_USDC`
ceiling so a pricing bug cannot drain the wallet, and reports the settlement tx and actual
spend.

## Checks

`yarn lint` clean on changed files, typecheck clean, and the existing x402 suites
(`routes/x402.test.ts`, `payments/x402-usdc-processor.test.ts`) pass 21/21.


vilenarios:
Thanks — all three findings were valid and are fixed in `1440ffb4`. One of them surfaced a regression I had introduced, which is worth calling out.

**Paywall vs facilitator credential (the TSDoc finding).** Documenting the wiring made it obvious that `cdpClientKey` fed *two* consumers: `createFacilitatorConfigFromCredentials` **and** `getPaywallHtml`. So my original fix, which repointed that field at the API key id, also pushed the id into the paywall markup where the public Client API key belongs. Rather than document the overload I split it: `cdpApiKeyId` for facilitator auth, `cdpClientKey` for the paywall. Neither can be substituted for the other now, which is a stronger guarantee than a comment.

**Shared rate-limit state.** Agreed — `x402-pay-test` depended on `x402-probe` having drained the bucket, and the probe deletes it on exit, so the dependency made the test order-sensitive and silently skippable. It now seeds and removes its own bucket. `x402-probe` is unchanged.

**Content-Length vs bytes received.** Also correct, and it was hiding a real defect. The harness now reads the stream and fails if the count disagrees with the header.

That last one led somewhere useful: draining the body was necessary but not sufficient. Settlement is submitted by the facilitator and confirms a block later, so the balance read still raced it and reported `spent: 0` on payments that had genuinely settled. It now waits for the settlement receipt and polls until the balance moves. Pinning the read to the settlement block was the obvious fix and does not work — public RPCs do not serve archive state.

Re-verified on Base mainnet after the changes:

```
402 -> sign -> retry -> 200
36,399,611 bytes received (== Content-Length)
settlement 0x1a228d2f1a18cc0378cb224db6ac34503ddbab2fc8a06a13bb2372bdffd224d3
confirmed block 50,221,387 · spent 0.004 USDC
```

Typecheck clean, 21/21 x402 tests passing.

vilenarios:
Fixed in `5a5e233e`, though the finding was half right and my own test is what made it look worse — worth correcting the record.

**What was genuinely wrong.** `createFacilitatorConfigFromCredentials` is exported and takes raw strings, so a direct caller passing `''` built a CDP config around an empty key id and earned a 401 instead of falling back. Blank is now treated as absent, whitespace included.

**What was not wrong: the config layer.** `env.varOrUndefined` already normalizes blank to undefined (`value.trim() !== ''`). So a compose-rendered `CDP_API_KEY_ID=` arrives as `undefined`, the URL fallback is selected, and the mainnet warning fires. Verified by loading config with all three CDP variables blank:

```
X_402_CDP_API_KEY_ID          = undefined
X_402_CDP_CLIENT_KEY          = undefined
X_402_CDP_FACILITATOR_KEY_ID  = undefined
[x402] MAINNET IS ENABLED WITHOUT COMPLETE CDP CREDENTIALS. ...
```

So blank credentials do not bypass the warning and do not produce a 401 through the normal path.

The review cited my test as evidence that they did, and that was fair — the test asserted exactly that. It claimed *"an operator who leaves CDP_API_KEY_ID blank therefore gets a CDP config built from an empty id and a 401"*, which skips the normalization happening before the helper is ever called. I had written it as a "documents real behaviour" case rather than checking whether the behaviour was real. It now asserts the correct behaviour instead.

Mutation-checked both guards: removing the blank normalization fails the blank-credential test, and reversing the credential precedence so the paywall key wins fails the wiring test. Neither guard is decorative.

28/28 passing, typecheck clean.

vilenarios:
Fixed in `37ee8624` — correct catch. The implementation normalizes both `cdpApiKeyId` and `cdpApiKeySecret`, but the test only exercised a whitespace-only key id, so a regression that stopped trimming the secret would have slipped through. Mutation-checked: normalizing only the key id now fails the test.

On the older `src/config.ts:3393` thread — that one is addressed in `5a5e233e`, with a correction to the premise. The config layer was never affected: `env.varOrUndefined` already normalizes blank to `undefined` (`value.trim() !== ''`), so a compose-rendered `CDP_API_KEY_ID=` arrives as undefined, the URL fallback is selected, and the mainnet warning fires. Verified by loading config with all three CDP variables blank. The helper genuinely did accept `''` and now doesn't.

That thread cited my test as evidence for the config-layer claim, and it was right to — the test asserted it, and the test was wrong. I'd written it as a "documents real behaviour" case without checking the behaviour was real. It now asserts correct behaviour instead.

28/28 passing, typecheck clean. Both guards mutation-checked: reversing the credential precedence fails the wiring test, and dropping either normalization fails the blank-credential test.

vilenarios:
Fixed in `0505f741` (+ lint cleanup in `1d6db51d`). This one found a real hole — the guard was not guarding anything in CI.

The wiring test asserted through the config module and branched on whatever `CDP_API_KEY_ID` happened to be. With neither variable set, which is exactly how CI runs, both sides of the comparison were `undefined` and the assertion held regardless of which credential won. I proved it: applying the precedence-reversal regression and running with the CDP variables unset passed **7/7**.

So my earlier "mutation-checked" claim was misleading. It only failed because I had those values exported in my shell. The test looked like a guard and was decorative in the only environment that matters.

The precedence is now extracted into `resolveFacilitatorKeyId()` and tested with explicit inputs, so it no longer depends on ambient environment. The config-level test only pins that the export is the resolver's output. There is also a separate case asserting the PUBLIC paywall key is never returned while a dedicated id exists, since that substitution is the security-relevant half of the bug and deserves its own assertion rather than being implied.

Re-checked under CI conditions — CDP variables unset, precedence reversed:

```
not ok 2 - resolveFacilitatorKeyId
# pass 9
# fail 2
```

Previously that same mutation passed silently.

32/32 tests, repo-wide `yarn lint:check` clean, typecheck clean. Also verified the startup warnings still fire after dropping the stale `eslint-disable` comments.

PR #870, opened 2026-08-20
feat(chunks): indexed eviction for the chunk data cache (ADR 005)

vilenarios:
Implements [ADR 005](https://github.com/ar-io/ar-io-node/blob/feat/chunk-data-cache-indexed-eviction/docs/madr/005-chunk-data-cache-indexed-eviction.md). Supersedes #866. Stacked on #867 (now merged).

The chunk cache is reclaimed only by `FsCleanupWorker`, whose walk spends **99% of each batch cycle traversing rather than deleting**. Reclaim loses to ingest and the volume has filled to 100% repeatedly. This adds the index-driven evictor the contiguous cache already has, keyed per `data_root` (~37.5k rows, *smaller* than the proven contiguous index).

**Everything is off by default and none of it has run in production yet.**

## Where to look first

The risk is concentrated in one place: **the delete path**, `src/workers/chunk-data-cache-evictor.ts` + `src/store/fs-chunk-data-store.ts#delDataRoot`. Everything else is schema, hooks, or docs.

Seven defects were found during review, all the same shape — *something that gates removal was incomplete, stale, or checked too early*. They are listed with their fixes in the ADR's Risks section. Two guards are what keep this safe now:

**1. A data root with an unconfirmed placement is never evicted.**
```sql
AND NOT EXISTS (SELECT 1 FROM chunk_placements p
                WHERE p.data_root = c.data_root AND p.confirmed_at IS NULL)
```
`chunk_placements` is the record of what is actually pending, so eviction asks it rather than inferring from a clock. An earlier version relied only on a config-derived age floor, which meant disabling the ingest cache — the natural response to disk pressure — shortened the floor out from under chunks still in flight. The floor remains as a backstop. Read-through chunks have no placement row and stay freely evictable, which is correct: they refetch.

**2. Reclaim is per file, against each file's own mtime — not `rm -rf` on the directory.** Anything the index thinks is cold but the filesystem says is fresh is kept. A directory-mtime check was tried and reverted: overwriting an existing offset bumps the *file's* mtime but never the directory's, so a re-POSTed chunk slipped through.

## Reviewing the rest

| Area | Files |
|---|---|
| Schema + DB methods | `migrations/*`, `sql/chunks/data-cache.sql`, `standalone-sqlite.ts`, `types.d.ts` |
| Write/read hooks | `store/fs-chunk-data-store.ts` |
| Workers | `workers/chunk-data-cache-{evictor,reconciler}.ts` |
| Wiring + config | `system.ts`, `app.ts`, `config.ts`, `metrics.ts` |
| Docs | ADR 005, `cache-cleanup.md`, `envs.md`, `glossary.md` |

## Verification

- **2,250 tests, 0 failures** (4 skipped, same as baseline). CI green on ubuntu and macOS.
- Every new test was checked against a deliberately reintroduced bug. **Five were rewritten after passing a mutation they should have caught** — the reason the count is trusted at all.
- **Measured against production data, read-only:** backfill over 128,969 data roots / 916,607 chunk files in **34.9s**; index-vs-filesystem accuracy **97.5%** on a random 400-row sample, every mismatch in the benign direction; the eviction query plans as an ordered index scan plus a PK-prefix probe against the real 912k-row `chunk_placements`.

Full measurements and methodology are in the ADR, including a sampling note: directory-count-weighted sampling under-estimated reclaimable bytes by nearly half, because the top 1% of data roots hold ~50% of chunks.

## Rollout

Nothing is enabled by default. `system.ts` double-gates the backfill so it cannot run without the index.

1. `ENABLE_CHUNK_DATA_CACHE_INDEX` — write hook only, evictor still off
2. `ENABLE_CHUNK_DATA_CACHE_INDEX_BACKFILL` — one pass, then unset
3. Evictor on one node; keep `FsCleanupWorker` as reconciler
4. Compare against the un-migrated node

Two operator notes, both in `cache-cleanup.md`: the watermarks default to `0`, which means the evictor never fires (it now warns at startup), and `FsCleanupWorker` does **not** honour the derived age floor — a pre-existing gap that matters more once both reclaimers run together.

## Deliberately not done

- **Write-hook coalescing.** A large upload issues one upsert per chunk against the same row, on the single write worker shared with `data.db`. Real, but a chunk-write hot-path change that deserves its own measurement. `CHUNK_DATA_CACHE_INDEX_UPDATE_ON_READ=false` is the zero-code mitigation.
- **Hybrid tail and ingest-origin tiering.** Knobs exist and are inert, per the ADR.

https://claude.ai/code/session_01Kfp8asuazZwWXZmReferQV


vilenarios:
## Adversarial review: **do not merge yet**

Three independent hostile reviews (data-loss paths, operational failure modes, test-quality mutation audit) were run against this branch. They found blocker-class issues that the existing tests do not catch. Posting in full rather than quietly fixing, because one of the findings invalidates a fix I pushed earlier today.

### Blocker 1 — index rows leak on every delete path except the evictor's

`deleteChunkDataCacheEntries` is the **only** thing that removes a row. Every other deleter on this box leaves the row behind:

- `FsChunkDataStore.del()` — called per chunk by `ChunkIngestGcWorker.evict`
- `FsCleanupWorker` for `chunk_data` — `ENABLE_CHUNK_DATA_CACHE_CLEANUP` defaults **true**, and `cache-cleanup.md` tells operators to keep it on as the reconciler
- the manual `find -delete` sweep this fleet has already used (778 GB)

`ChunkDataCacheIndex` has no single-row delete at all, so the plumbing does not exist. Phantom rows are the **coldest** rows — `last_access` freezes when the files vanish — so `ORDER BY tier ASC, last_access ASC` sorts them to the **front of every batch**. `delDataRoot` treats ENOENT as success and bytes are booked against the deleted row, so the first pressure sweep can report ~50 GiB freed across 50k phantoms while `df` does not move. `ON CONFLICT DO NOTHING` means the backfill cannot heal it.

This is the gw1 failure ADR 005 cites — except the drift warning never fires, because candidates are never empty. **The one diagnostic built for this class of bug is bypassed by it.**

### Blocker 2 — stock watermarks are `0`, so the evictor never fires

`CHUNK_DATA_CACHE_HIGH_WATERMARK_PERCENT`, `_LOW_`, `_MIN_FREE_BYTES` all default to 0, making `overPressure()` unconditionally false. Enabling the master flag on a stock node buys both hooks, unbounded table growth and WAL churn, and evicts nothing. Config *throws* on an inverted watermark pair but says nothing here.

### Blocker 3 — `UNLINK_CONCURRENCY = 50` vs `UV_THREADPOOL_SIZE` default 4

50 concurrent recursive `fs.rm` calls take the entire libuv pool on a stock node, queueing every chunk read behind them on a device that is by definition saturated when the evictor runs. `warnIfWalkConcurrencyUnsafe` cannot see it — it only sums `FsCleanupWorker` instances, so neither the evictor nor either reconciler is counted.

### The fix I pushed earlier is wrong — reverted

I added a directory-mtime guard to `delDataRoot` as a last line of defence against a failed index write leaving `last_write` stale. It is **incomplete and net-harmful**, both verified on the box:

- **Overwriting an existing offset does not bump directory mtime.** A re-POSTed chunk at an offset already cached is exactly an ingest path, and the guard is blind to it. I had actually measured this and reasoned it away as "not the propagation risk". That was wrong.
- **Unlinking a chunk *does* bump directory mtime.** `FsCleanupWorker` and the ingest GC unlink constantly, so the guard would veto evictions routinely — *after* the row was already deleted, manufacturing orphans and stalling reclaim.

Reverted; the branch is back to the state CI validated. The underlying hole is real and now tracked below.

### Also found

- **DELETE → `rm` gap is unguarded.** A chunk landing there is destroyed with no index-write failure required. A stat-based check cannot close it; the sound fix is `rename()` to a tombstone first, then inspect, then remove.
- **Disabling `CHUNK_INGEST_CACHE_ENABLED` collapses the floor** from 24h to 1h at module load, over chunks already on disk with hours of confirmation window left — and disabling it is the natural response to disk pressure.
- **No coalescing on the write hook.** A 25,000-chunk upload issues 25,000 separate upserts against *one row*, each its own transaction with `synchronous=FULL`, on the **single** write worker shared with data.db. The queue is uncapped.
- **`flushFailed` is per-run, not per-shard** (my earlier fix): one transient `SQLITE_BUSY` disables checkpointing for the remaining ~4000 shards. `queueWrite` is called with `maxRetries = 0`.
- **The WAL premise is wrong and the fix has a cost.** `wal_autocheckpoint` is never disabled, so chunks.db-wal does not grow unbounded. Meanwhile `wal_checkpoint(TRUNCATE)` runs on the shared `data` pool with 30+ live connections and a 30s busy timeout — a potential head-of-line stall of the data.db writer every 15 minutes.
- **The backfill checkpoint lives under `data/chunks`**, which `FsCleanupWorker` walks and deletes at 4h. Pause a backfill overnight and it re-walks the whole tree.
- **`chunk_cache_index_skipped_floor_total` is structurally always zero** — it only increments if the SQL floor stops working. Three docs claim it shows rows held back by the floor, and the rollout tells operators to watch it. It should be a gauge of `COUNT(*) WHERE last_write > cutoff`.
- **Every index and unlink failure logs at `debug`**; this box runs `info`. The feature can fail 100% silently.
- `ENABLE_CHUNKS_DB_WAL_CLEANUP` is missing from `docs/envs.md`, and ADR 005 still lists the WAL risk as open although the fix is in this branch.

### Test gaps (mutation audit, 100 tests)

Mutations that **survived**: removing the `sweeping` re-entrancy guard; removing the `minFreeBytes` pressure trigger (every test passes `minFreeBytes: 0`); unconditional `clearCheckpoint`; the `cleanupWal` pool-routing line — the one genuinely new piece of logic there. No test imports `metrics`, so the whole `chunk_cache_index_*` family is unverified. The integration age-floor test passes with two of its three guards deleted.

### Assessment

The age-floor logic itself held up — the SELECT/in-memory/DELETE triple was attacked directly and the select→delete race is genuinely closed. What did not hold up is everything around it: lifecycle, coexistence with the other two reclaimers, and observability. Blocker 1 in particular means the feature would under-deliver silently in exactly the incident it exists to prevent.

Recommend this returns to draft until at least Blockers 1–3 are resolved.

vilenarios:
### Status: the blockers above are resolved

Superseding my earlier "do not merge yet" comment so it does not read as a standing objection. Every blocker it raised is fixed and covered by a test verified to fail without its fix.

| Finding | Resolution |
|---|---|
| Index rows outlive their files; evictor reports bytes `df` never shows | Reclaimed bytes now come from the filesystem, not the index row. Stale rows still get purged, so the index self-heals, and they are counted on their own series |
| Stock watermarks are `0`, so the evictor never fires | Warns at startup when no pressure trigger is configured |
| `UNLINK_CONCURRENCY = 50` vs a default libuv pool of 4 | Derived: `max(1, UV_THREADPOOL_SIZE / 8)` |
| DELETE → `rm` gap destroys a chunk landing in it | Reclaim is per file against each file's own mtime; anything not provably old is kept |
| Disabling the ingest cache collapses the age floor | Eviction now excludes any data root with an unconfirmed placement; the floor is a backstop |
| Backfill checkpoint sat inside the tree `FsCleanupWorker` sweeps | Moved to `data/tmp/` |
| One dropped batch disabled checkpointing for the whole pass | Scoped per shard |

Two decisions worth surfacing rather than burying:

**The `chunks.db` WAL worker was removed, not kept.** It was written against a claim in my own ADR revision that the WAL grows unbounded. Measurement says otherwise — `wal_autocheckpoint` is never disabled, and the file sits at 11,857,392 bytes, byte-identical across a 30s sample and unchanged over 19 hours. It would have reclaimed ~12 MB while running an exclusive `TRUNCATE` checkpoint on the shared `data` pool against 30+ live connections. `cleanupWal` is now byte-identical to `develop`.

**A directory-mtime guard was tried and reverted.** Overwriting an existing offset bumps the *file's* mtime but never the directory's, so a re-POSTed chunk slipped through it — and because unlinking *does* bump directory mtime, it would have vetoed evictions constantly. The per-file check has neither flaw.

**Write-hook coalescing is deliberately not in this PR.** It is a chunk-write hot-path change and deserves its own measurement; `CHUNK_DATA_CACHE_INDEX_UPDATE_ON_READ=false` is the zero-code mitigation if the write load bites.

Still worth a careful human read of the delete path. Eight defects of one shape were found here — something gating removal was incomplete, stale, or checked too early — three by adversarial review, two by CodeRabbit, and one introduced while fixing another. The tests catch all eight now, but that rate is the argument for a second pair of eyes rather than against it.

PR #878, opened 2026-09-03
feat(chunk): default chunk retrieval to miners before gateway peers

vilenarios:
## What changed

`CHUNK_DATA_RETRIEVAL_ORDER` and `CHUNK_METADATA_RETRIEVAL_ORDER` now default to `arweave-network,ar-io-network` instead of `ar-io-network,arweave-network`. `docs/envs.md` gains rows for both, plus the two `*_SOURCE_PARALLELISM` knobs, none of which were documented. `docker-compose.yaml` already passes all four through, so it needs no change.

## What needs your judgement

1. **This ships on a load argument alone, with no delivery claim.** The measurement could not resolve a delivery effect (below). If you want a delivery claim before merging, this needs days of data, not hours.
2. **One interaction is real and is called out below**: abandoned requests move from the cancellable path to the non cancellable one.
3. **Peer discovery warmup** might deserve parallel work. This change makes it matter more.

## Why

Arweave miners are incentivized to store and serve arbitrary chunks. AR.IO gateways only hold what they happened to cache. Asking the gateway layer first therefore turns the common case into a near certain miss, and those misses concentrate on whichever gateway originated the data.

Probing four independent AR.IO gateways for an offset that a node serves successfully returned clean 404s from all four. The peer tier is healthy. It just is not where arbitrary chunks live.

## Evidence, and its limits

Measured on **one development gateway**, in paired alternating windows:

- Outbound AR.IO peer chunk requests fell **23 to 41 percent** at equal request volume. This is mechanical: the peer tier only ever sees what the miner path missed, so it generalises.
- **No delivery claim.** Peers first serves `|P| + |M\P|`, miners first serves `|M| + |P\M|`, which are the same union, so ordering changes who serves and what it costs rather than whether a chunk is obtainable. Three windows of the *same* config spanned 9x in measured delivery rate, so the metric cannot resolve an effect this size at this duration. Both directions appeared in the data depending on the window.

## Risks

1. **Load shifts to the Arweave node set**, and for nodes that have not discovered many peers, onto `TRUSTED_NODE_URL` (`arweave.net` by default). Bucket filtered peer selection left about 10 candidate nodes per offset on a gateway that knew 128.
2. **Abandoned requests.** When a client disconnects, the peer fan out cancels, because the client signal is merged into each peer fetch. The miner fetch does not: `getChunkByAny` deliberately keeps the shared `chunkPromiseCache` entry running so other waiters still receive the result. Miners first therefore moves that work to the non cancellable path. It still populates the cache, and measured total outbound volume fell rather than rose, but the interaction is real.

## Rollback

Set either variable back in `.env` and restart. Operators who already tuned these are unaffected: only the default moves. No migration, no schema change, and nothing here touches verification or the trust headers.


vilenarios:
The mechanism is real, one of the numbers is not, and I do not think the fix belongs in this PR.

**Confirmed.** The shared fetch is uncancellable. `chunkPromiseCache`'s read-through function calls `peerGetChunk` with no signal (`composite-client.ts:359-371`), and `getChunkByAny` only races the caller's await against the abort. That is deliberate, so other waiters still get the result, and this PR's body already names it as risk 2.

**Corrected.** The bound is not 50 peers. `peerGetChunk` loops `Math.max(peerSelectionCount, retryCount)` times, and production wiring passes `chunkGetPeerSelectionCount = ARWEAVE_PEER_CHUNK_GET_PEER_SELECTION_COUNT` (default 10) and `chunkGetRetryCount = ARWEAVE_PEER_CHUNK_GET_MAX_PEER_ATTEMPT_COUNT` (default 5) (`system.ts:276-278`). So the real bound is **10 sequential attempts at `PEER_CHUNK_REQUEST_TIMEOUT_MS = 500`, about 5 seconds**, not 25. The 50 is the constructor default, which no deployment uses.

**Scope.** This PR flips two default strings. It does not create the detached fetch, and the fetch behaves identically with either ordering. What changes is how often abandoned requests land on that path, which is worth saying and is said. Adding a cleanup bound plus delayed-miner tests for both the data and metadata entry points is a behavioral change to `ArweaveCompositeClient` and wants its own review, not a rider on a default flip.

Measured for context, since it decides whether this is urgent: over a 146-hour window on a production gateway, 99.3% of peer-origin chunk requests past the cache were cancelled by the caller, and 0.31% delivered bytes. So detached work is the common case rather than the exception, but it is bounded at roughly 5 seconds and it populates the cache, which is why it was built this way.

Filed separately so it is not lost: #885.

Issue #879, opened 2026-09-03
Chunk metadata anchoring probes a single default gateway for the whole network


vilenarios:
## Summary

`ChunkMetadataAnchorSource` resolves `offset → tx + data_root` by sending `HEAD /chunk/{offset}/data` to a reference peer and reading the `X-Arweave-Chunk-*` headers. Its peer pool is `GATEWAYS_ROOT_TX_URLS`, which defaults to a single gateway. Every node running defaults therefore probes the same host, and the volume scales with the size of the network rather than with that host's own traffic.

Enabled by default since r79 via `CHUNK_METADATA_ANCHOR_ENABLED`.

## Mechanism

- `HEAD /chunk/{offset}/data`, falling back to a zero byte range `GET` on the same path when HEAD returns no usable headers: `src/data/chunk-metadata-anchor-source.ts:277`
- Peer pool resolution, which reuses the root tx pool: `src/system.ts:986`
- Default pool, one entry: `src/config.ts:666`
- `validateStatus: () => true` (`:148`), so a 404 is not an error and nothing backs off
- LRU keyed by tx id, 1024 entries and a 300s TTL by default, so a stream of distinct offsets rarely caches
- No hop guard. `validateHopCount` is enforced in `ar-io-chunk-source.ts`, `ar-io-data-source.ts` and `gateways-data-source.ts`, but not here, so a peer origin request can trigger an outbound probe

## Measured rate

One gateway on defaults, 10.2 hours, `ario_chunk_metadata_anchor_total`:

| result | count |
|---|---:|
| `hit` | 1,351 |
| `cache_hit` | 2,827 |
| `metadata_missing` | 9,865 |
| `error` | 7 |

Cache hits cost nothing, so that is **11,223 network lookups, about 26k per node per day**. Across a network of several hundred gateways that converges on one host at a rate in the millions of requests per day.

The observer carries the same counter (`observer_chunk_metadata_anchor_total`), so a deployment can have two emitters per node. It recorded no lookups on the node measured here, so 26k/day is a floor rather than a total.

## This is not the `ar-io-network` chunk source

`ArIOChunkSource` uses undici `fetch` and requests `GET /chunk/<offset>` with no `/data` suffix (`src/data/ar-io-chunk-source.ts:502`). It cannot produce an axios user agent or the `/data` path. The anchor source is the only code in this repo that emits `HEAD /chunk/<offset>/data`.

[...]

PR #880, opened 2026-09-03
feat(chunk): shadow and enforce modes for peer-origin chunk serving

vilenarios:
Addresses option 4 in #879.

## What changed

`CHUNK_PEER_ORIGIN_MODE`, default **off**. It governs chunk requests arriving with `X-AR-IO-Hops >= 1`, meaning another AR.IO gateway forwarded them.

| mode | behavior |
|---|---|
| `off` | unchanged: the full cascade runs for peers |
| `audit` | unchanged, plus a record of what enforcing would have cost |
| `enforce` | answer from sources reachable without the network, else the usual not-found 404 |

An invalid value fails startup. The naming follows Kubernetes PodSecurity admission (`enforce` / `audit`), the same idea as SELinux permissive, AppArmor complain and CSP report-only.

## Why

A peer allows **one second** before it gives up (`PEER_REQUEST_TIMEOUT_MS`, `ar-io-chunk-source.ts:42`), while `CHUNK_SERVE_DEADLINE_MS` bounds our own serve at 12s by default, a figure sized against the upstream proxy cut rather than a peer's patience. Remote work started for such a caller outlives it, and that caller runs its own cascade anyway.

The AR.IO peer sources already bound forwarding depth with `validateHopCount` (`MAX_CHUNK_HOPS = 1`), so peer-origin requests do not re-fan-out to peers. The boundary lookup, the anchor probe and the Arweave node path carry no such bound.

Measured on a development gateway over **14.72 hours in `audit`**, with behavior unchanged:

| | |
|---|---:|
| peer-origin requests past the cache | **82,566** (93.5/min) |
| cancelled before completion | 81,381 (**98.6%**) |
| delivered bytes | **174** (0.21%) |
| served from cache, which `enforce` keeps | 1,320 |
| outbound requests spent on them | **130,060** (147/min) |
| **outbound per chunk delivered to a peer** | **747** |

Split of that outbound: 80,533 AR.IO peer attempts, 31,727 chunk-metadata anchor probes, 17,800 Arweave node requests. On this gateway `enforce` would trade **8,836 outbound requests per hour against 11.8 deliveries per hour**.

## `enforce` declines by source, not by pipeline position

This is the part worth reviewing closely.

Stopping at the cache lookup would be wrong on a gateway that originates data. A chunk cached at ingest is stored by data root and relative offset (`ingest-chunk-cache.ts:192-193`), because a chunk posted before mining has no weave offset yet, so no absolute-offset symlink is written. `tryCacheHit` reads by absolute offset alone (`fs-chunk-data-store.ts:171`), so an ingested chunk misses it and is reachable only once the local index resolves its boundary. Refusing everything past the cache would refuse peers exactly the chunks only that gateway holds.

So `enforce` runs the pipeline with every network source declined:

- `CompositeTxBoundarySource` skips the anchor probe, tx_path validation and the chain fallback, keeping the database source, which needs no network.
- `ArweaveCompositeClient.getChunkByAny` declines, covering `getChunkDataByAny` and `getChunkMetadataByAny` since both funnel through it.
- `ArIOChunkSource` needs no change: it already honors `skipRemoteForwarding`.
- `ReadThroughChunkDataCache` still reads local disk by `(dataRoot, relativeOffset)` first, which is how the ingested chunk is served.

`localSourcesOnly` is a new request attribute rather than a widening of `skipRemoteForwarding`, whose meaning stays "skip the AR.IO peer layer" for compute-origin callers that may still reach Arweave nodes. `TxBoundarySource.getTxBoundary` gains an optional third parameter to carry attributes; additive, and existing implementations ignore it.

`TxBoundary` also gains an optional `source` (`db`, `anchor`, `tx_path`, `chain`) set by the composite, without which `audit` cannot tell local resolution from remote.

## What `audit` is for

`chunk_peer_origin_audit_total{boundary,bytes}` records, per peer-origin request that missed the cache, whether offset resolution stayed local and whether the bytes came off local disk.

**Read the `bytes="local"` cells: they are the cost of enforcing.** Both mean this gateway held the chunk and served it.

| cell | 14.7h window | under `enforce` |
|---|---:|---|
| `boundary="local"`, `bytes="local"` | 0 | still served |
| `boundary="remote"`, `bytes="local"` | 29 | **no longer served**: the bytes were locatable only through a network offset lookup |
| `boundary="remote"`, `bytes="remote"` | 145 | no longer served, proxied from the network |

An earlier revision of this PR told operators to check `boundary="local",bytes="local"` alone. That is too narrow, and the soak is what caught it: that cell was 0 while the real cost was 29. The guidance in `envs.md` and both code comments are corrected.

Run `audit` and read those cells before enabling `enforce` anywhere that originates data. On the gateway used here the total is 29 in 82,566 requests, 0.035%, and `chunk_ingest_pending_bytes` stayed 0 throughout, which is expected: that node is not an origin, so it never exercised the ingest case. That is exactly why the number has to be measured per gateway rather than argued from one box.

## Verified end to end

Built, deployed to a development gateway and exercised against the live node.

`enforce`, earlier revision:

| leg | request | result |
|---|---|---|
| peer-origin, offset in local cache | `X-AR-IO-Hops: 1` | 200 in **13ms** |
| peer-origin, offset not cached | `X-AR-IO-Hops: 1` | 404 in **11ms**, with the `ar-io-network`, `getChunkByAny` and anchor counters **unmoved** |
| same offset, no hop header | none | 404 after **3,468ms**, `getChunkByAny` +3 |

The third leg is the control: without the header the cascade still runs, so the policy is scoped to peer traffic rather than suppressing retrieval generally.

`enforce`, current revision, which declines by source rather than stopping at the cache:

| leg | result |
|---|---|
| peer-origin, nothing local can resolve it | 404 in **12ms**, with the `ar-io-network`, Arweave node and anchor counters **all unmoved** |
| same offset, no hop header | 404 after **3,204ms**, Arweave node counter +2 |

`audit`, same node and request shape: 404 after **3,338ms**, matching `off` rather than `enforce`, with the outcome recorded.

One property is covered by unit test rather than live demonstration. `ReadThroughChunkDataCache` writes to the store *with* the absolute offset (`read-through-chunk-data-cache.ts:152-157`), so any chunk this gateway fetched is reachable through `tryCacheHit`. Only the ingest path writes without one. The "serve a chunk the local index resolves" case therefore only appears on a gateway that receives uploads, and the dev gateway used here had `chunk_ingest_pending_bytes` at 0 for the whole soak.

Unit tests cover the cache hit, the flags reaching both the boundary source and the chunk source, a locally indexed chunk still being served under `enforce`, refusal when no local source can serve, `hops=0`, absent attributes, the disabled default, and the audit outcomes. `composite-tx-boundary-source.test.ts` is new: the file had no coverage, and both the audit labels and the new local-only gate depend on it. 38 tests across the two files.

## Limits

- Detection depends on the caller sending `X-AR-IO-Hops`. Our nodes do, via `generateRequestAttributes`. Any other client reads as `hops=0` and is unaffected, so this bounds fleet traffic rather than arbitrary clients.
- `enforce` means peers relying on a gateway to proxy chunk fetches receive 404s for anything it cannot serve locally. That is the intent, and it is why the default is `off`.
- Operator-owned storage backends such as `legacy-s3` are treated as local and are not declined. The rule denies the network, not the operator's own storage.

## Test ladder

`eslint src test` clean. `data-verification.test.ts` reports 3 failures when the suite runs on my machine, on this branch and on clean `develop` at `7466d916` alike, so they are not from this change; CI is green on #878 from the same base.


vilenarios:
Both CodeRabbit findings are addressed, and the mode has been renamed.

**Formatting (`chunk-retrieval-service.ts:301`)**: correct, and it was the CI failure. Six `prettier/prettier` errors from mechanical edits I made after my last lint run. `eslint src test`, which is what CI runs, is now clean.

**Legacy `CHUNK_PEER_ORIGIN_LOCAL_ONLY` (`docker-compose.yaml:178`)**: no migration path is needed, because that variable never existed outside this branch. It was introduced and replaced in the same unmerged PR, so no released version reads it and no deployment can be relying on it. Adding a fallback would ship a deprecated alias for a variable that was never public. The one machine that had it set was mine, and it is migrated.

**Rename `shadow` to `audit`**: raised separately, and right. "Shadow" in infrastructure usually means mirroring traffic to a parallel stack, which this does not do. It evaluates a policy and records the verdict without acting on it, which is `audit` in Kubernetes PodSecurity admission (`enforce`/`audit`), permissive in SELinux, complain in AppArmor, report-only in CSP. The metric is now `chunk_peer_origin_audit_total`.

**`enforce` no longer stops at the cache**, which is the substantive change since the last review. It runs the pipeline with every network source declined instead. The old shape refused anything past `tryCacheHit`, which reads by absolute offset only, so it would have refused chunks the gateway holds: one cached at ingest is stored by data root and relative offset, since a chunk posted before mining has no weave offset yet. On a gateway where uploads land, that meant refusing peers exactly the chunks only it had. Declining by source rather than by pipeline position keeps the local index and local chunk store in play and still does no network work.

vilenarios:
Both new findings were valid and are fixed in `b45379f1`.

**`legacy-s3` audited as remote.** Correct, and worse than a label problem: it made the audit disagree with `enforce`. `enforce` permits operator-owned storage, so recording an S3-served chunk as `bytes="remote"` overstated the cost of enforcing. Classification now uses a named `LOCAL_CHUNK_SOURCES` set (`cache`, `legacy-s3`) with a comment tying it to what `enforce` declines, so the two cannot drift apart silently. Test added for an S3-sourced chunk landing in `bytes="local"`.

**Local infrastructure errors reported as not-found.** Correct, and this one mattered. Under `enforce` every non-cancellation error was wrapped as `peer_origin_local_only`, which the route maps to 404, so a SQLite or disk failure looked like "this offset does not exist". Fixed in two places:

- `CompositeTxBoundarySource` now propagates a database error when `localSourcesOnly` is set. In that mode there is no fallback, so the error is the whole answer rather than one miss among several.
- Only a `ChunkNotFoundError` is wrapped as a local-only refusal. Anything else keeps its own error so `classifyChunkRetrievalError` can report a gateway fault.

Test added: a throwing database source surfaces as an infrastructure error rather than a refusal.

The two earlier findings stand as previously answered: formatting is fixed and `eslint src test` is clean, and the legacy `CHUNK_PEER_ORIGIN_LOCAL_ONLY` needs no compatibility path because it was introduced and replaced inside this unmerged PR, so no release reads it.

PR #881, opened 2026-09-03
fix(chunk): classify an internal abort as 404, not 502

vilenarios:
## What changed

An `AbortError` reaching `classifyChunkRetrievalError` now returns **404 `upstream_aborted`** instead of falling through to the 502 default.

## Why

A client disconnect is already handled in the branch above, which returns 499. So an `AbortError` that reaches the rest of the classifier is an **internal** cancellation: a source's own abort signal, or a losing peer cancelled once another peer won the `Promise.any` race in `ar-io-chunk-source.ts`.

Reporting that as 502 contradicts the contract the wall-clock deadline was built around, stated in `handlers.ts` itself: a chunk fetch that fails for any reason is a 404, which keeps slow and failed serves out of the 5xx count, lets the cascade fall through, and lets nginx serve a cached negative. An internal cancellation is the same condition as a timeout, which already maps to 404.

## Evidence

Observed on a live gateway:

```
warn: Unable to retrieve chunk from upstream sources This operation was aborted
  errorType: "upstream_unavailable"  statusCode: 502  offset: 383419471572868
```

`This operation was aborted` is the `AbortError` message. In one 3-minute window, **22 of 292** inbound `/chunk/:offset` requests were reported as 502 this way, and the logs carried 28 occurrences in 25 minutes. The same offset repeats across requests, so peers retry against a status that is neither cacheable as a negative nor accurate.

## Verification

The new test reproduces the bug before it fixes it:

```
$ git checkout develop -- src/routes/chunk/handlers.ts
not ok 1 - maps an internal AbortError to 404 upstream_aborted, not 502
  expected: 404
  actual: 502
# pass 4  # fail 1

$ git checkout HEAD -- src/routes/chunk/handlers.ts
# pass 5  # fail 0
```

The suite also asserts the paths this must not disturb: a client-aborted request still returns 499, `ChunkServeTimeoutError` still returns `serve_deadline_exceeded` so it stays attributable to `chunk_serve_deadline_exceeded_total`, a genuine upstream failure still returns 502, and `ChunkNotFoundError` keeps its own `errorType`.

## Note on the test file

The cases are in a new `classify-chunk-retrieval-error.test.ts` rather than appended to `handlers.test.ts`, because that file is owned by root in some working trees and cannot be written without a chown. Nothing is duplicated between the two files.

### DIFF
--- a/src/routes/chunk/classify-chunk-retrieval-error.test.ts
+++ b/src/routes/chunk/classify-chunk-retrieval-error.test.ts
@@ -1,0 +1,81 @@
+/**
+ * AR.IO Gateway
+ * Copyright (C) 2022-2025 Permanent Data Solutions, Inc. All Rights Reserved.
+ *
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+import { strict as assert } from 'node:assert';
+import { describe, it } from 'node:test';
+
+import { ChunkNotFoundError } from '../../data/chunk-retrieval-service.js';
+import {
+  ChunkServeTimeoutError,
+  classifyChunkRetrievalError,
+} from './handlers.js';
+
+/**
+ * Cancellation cases for `classifyChunkRetrievalError`.
+ *
+ * These live here rather than in `handlers.test.ts` because that file is not
+ * writable in every working tree. The cases in `handlers.test.ts` still cover
+ * the timeout, not-found and unrecognized-failure paths; nothing is
+ * duplicated between the two.
+ */
+describe('classifyChunkRetrievalError - cancellation', () => {
+  const abortError = () =>
+    Object.assign(new Error('This operation was aborted'), {
+      name: 'AbortError',
+    });
+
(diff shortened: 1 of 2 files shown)


vilenarios:
Declining this one, with evidence.

The guideline reads "Always use `createTestLogger()` from `test/test-logger.ts` in test files; never `winston.createLogger({ silent: true })`." Its purpose is to forbid ad-hoc winston loggers that write to the console, not to require a logger in a test that has none.

`classifyChunkRetrievalError` is a pure function: it takes an error and a boolean and returns `{statusCode, errorType}`. The test constructs no logger of any kind, so there is nothing for the rule to prevent.

That also matches how the repo is actually written. 53 of 128 test files have no `createTestLogger`, and they are exactly the pure-function suites: `config.test.ts`, `filters.test.ts`, and all of `src/lib/`. Adding an unused logger here would be the outlier.

If the intent is that every test file must construct one regardless, that belongs in the guideline as an explicit rule and would want a lint check behind it rather than a per-PR comment.

